PR #6, opened 2019-07-13
update version

thery:


maximedenes:
Something is wrong in the included commits, but anyway I'll update the readme.

PR #40, opened 2019-08-06
[parser] understand lp:{{...}} quotations

gares:


gares:
Can I do something here? I've been using this patch for some time and it works for me ;-)

PR #56, opened 2019-09-18
Change keybindings for macOS to avoid beeps

TheoWinterhalter:
I'm not sure these are the optimal choices, but this is a proposal to change the keybindings as the current ones trigger annoying beeps.


maximedenes:
Looks good (I'm not very familiar with macOS, though, so I can't comment on how optimal the new bindings are). Thanks for contributing!

PR #59, opened 2019-09-20
Output spaces instead of non-breaking spaces

### LINKED ISSUE #58: Output spaces instead of non-breaking spaces
This is the issue purportedly fixed in https://github.com/siegebell/vscoq/pull/164 (haven't tested the fix in this version).

To test this, try pasting output from any command into a Coq buffer. The parsing errors you get are very hard to debug.
```coq
Lemma plusnO : forall n : nat, n + 0 = n. Admitted.
Check plusnO.
(* Paste the output for a new lemma. *)
Lemma plusnO1 : forall n : nat, n + 0 = n.
(* Error: Syntax error: '.' expected after [vernac:gallina] (in [vernac_aux]). *)
```

Here's a hexdump of the offending line. The non-breaking spaces are visible as (0x)`c2 a0`, while spaces are (0x)`20`.
```
$ echo 'Lemma plusnO1 : forall n : nat, n + 0 = n.'|hexdump -C
00000000  4c 65 6d 6d 61 20 70 6c  75 73 6e 4f 31 20 3a c2  |Lemma plusnO1 :.|
00000010  a0 66 6f 72 61 6c 6c c2  a0 6e c2 a0 3a c2 a0 6e  |.forall..n..:..n|
00000020  61 74 2c c2 a0 6e c2 a0  2b c2 a0 30 c2 a0 3d c2  |at,..n..+..0..=.|
00000030  a0 6e 2e 0a                                       |.n..|
00000034
```

Blaisorblade:
A rebase of @varkor's https://github.com/siegebell/vscoq/pull/164. Fixes #58.

The code is all @varkor; I only rebased it and tested it on the example in #58 (I had tested the original code before).

The fix seems reasonable to me, tho I'm no JS expert.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * fix #26 (Focus lost on first go to point)
 * fix #33 (Top module name should be set properly)
 * fix #39 (UnhandledPromiseRejectionWarning)
+* fix #59 (Output spaces instead of non-breaking spaces)
 * fix #60 (Clear the goals when not in proof mode)
 
 Note: PR and issue numbers below refer to the previous VsCoq repository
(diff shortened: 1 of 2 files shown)


maximedenes:
Isn't this a Coq bug, though?

Blaisorblade:
Maybe, or maybe there's some complex reason for it. Either way, this fix works with existing Coq versions (and I guess would work with future ones that emitted spaces), it's very little code, and the bug could be pretty nasty (tho not for everybody).

maximedenes:
Thanks!

PR #64, opened 2019-09-28
Single build entry point.

vlj:
This PR adds a package.json file in the root directory of the project which is responsible for dispatching calls to client/server/html_views ; it also merge the 3 .vscode subfolder inside the root directory of the project.

This allows the following workflow of development :
- open folder > root directory 
- ctrl + shift + B to launch a build of everything
- F5 to launch a debug session of the client code (with breakpoint and everything else)

I tested the workflow on Windows but it should work on all platform (nothing OS specific here).
The benefit over current situation is that it's possible to have a look at the whole solution and not just the client subfolder to be able to debug, and the ability to debug client and server (there is a small caveat though, CoqLanguageServer.ts needs to be modified so that the debug options are passed even in "run" mode to be able to attach to server) inside the same vscode session.
Another benefit is that devDependencies are shared between the 3 subfolders.


.gitignore
```diff
@@ -1,3 +1,9 @@
+node_modules/
+client/out
+server/out
+html_views/out
+package-lock.json
```
maximedenes: Not sure about this one. Shouldn't it be versioned, as npm says sometimes?

.vscode/launch.json
```diff
@@ -0,0 +1,46 @@
...
+		{
+			"type": "node",
+			"request": "attach",
+			"protocol": "inspector",
+			"name": "Attach to Server",
+			"port": 6005,
+			"restart": true,
+			"outFiles": ["${workspaceRoot}/server/out/**/*.js"]
+        }
+        // ,
```
maximedenes: Is this a leftover from a previous configuration? If so, it should be removed.

package.json
```diff
@@ -0,0 +1,45 @@
+{
+	"name": "vscoq",
+	"displayName": "VSCoq",
+	"description": "An IDE for the Coq Proof Assistant",
+	"author": {
+		"name": "Christian J. Bell",
+		"email": "siegebell@gmail.com",
+		"url": "http://people.csail.mit.edu/cj/"
+	},
+	"version": "0.2.8",
```
maximedenes: The version in `master` is now 0.3.0. It was probably updated after you started working on this PR.

maximedenes:
The proofview problem is indeed fixed, thanks!

I was wondering: shouldn't the extension-related configuration (`contributes` and the likes) be moved from `client/package.json` to `package.json`? That's what https://github.com/microsoft/vscode-eslint does, it seems.

Then, `vsce` should probably be installed by the root `package.json` and run there. Also, we could finally get rid of the duplicated `README.md` and `CHANGELOG.md` in `client`.

What do you think?

PR #66, opened 2019-10-03
Use `coqidetop.opt` instead of `coqidetop`

### LINKED ISSUE #65: VsCoq does not work with Coq built using Dune
Reported by @mattam82. The problem is that dune builds `coqidetop.opt`, but not `coqidetop`. I imagine we could always use the `.opt` variant.

maximedenes:
Fixes #65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * fix #39 (UnhandledPromiseRejectionWarning)
 * fix #59 (Output spaces instead of non-breaking spaces)
 * fix #60 (Clear the goals when not in proof mode)
+* fix #65 (VsCoq does not work with Coq built using Dune)
 
 Note: PR and issue numbers below refer to the previous VsCoq repository
 (https://github.com/siegebell/vscoq).
--- a/server/src/coqtop/CoqTop8.ts
+++ b/server/src/coqtop/CoqTop8.ts
@@ -204,7 +204,7 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
   }
 
   private get coqidetopBin() {
-    return path.join(this.settings.binPath.trim(), 'coqidetop');
+    return path.join(this.settings.binPath.trim(), 'coqidetop.opt');
   }
 
   private spawnCoqTop(mainAddr : string, controlAddr: string) {


gares:
Not so sure this is the right fix. The dune file deliberately chooses the .opt extension, while the Makefile gives you an "extensionless" variant which is the one to be used by default (IIRC it is called BEST internally). See:
https://github.com/coq/coq/blob/master/ide/dune#L19-L20
I think it would make sense to ask *why* the dune file decided to do things this way..

gares:
I hereby coin the term "flash PR" ;-)

maximedenes:
> while the Makefile gives you an "extensionless" variant which is the one to be used by default (IIRC it is called BEST internally).

I'd rather ask: Why did make go this way?

gares:
I imagine becuase the entire world does not put in the binary name the name of the compiler backend, but it was not me making the choice, so I don't know.

gares:
I mean, as a final user, I'd rather type `foo` and not `foo.gccO3`...

maximedenes:
> I mean, as a final user, I'd rather type `foo` and not `foo.gccO3`...

Sure, but as a final user, you'll never launch `coqidetop`, will you?

On a higher-level, I'm not sure what you see wrong with this patch. Of course, the question can be discussed on the Coq side, but VsCoq is not going to change the name of executables, and we want it to work with the current version of Coq so I'm not sure what else than this patch we could do, TBH.

gares:
More seriously, this change means that if you `make byte` then vscoq cannot pick up the toplevel. I don't think it is a big problem for final users, but it may hit devs. FTR, I'm not suggesting making the code clever and pick one among .opt .byte .exe ... I'm just highlighting a potential issue.

gares:
(Also, there are some platforms where .opt cannot exist)

maximedenes:
> More seriously, this change means that if you `make byte` then vscoq cannot pick up the toplevel. I don't think it is a big problem for final users, but it may hit devs.

That's a good point, but I guess we should rather discuss it in a Coq-side issue, don't you think? As long as Coq maintains two build systems, any workflow difference between the two will make the life of external tool developers more painful.

> (Also, there are some platforms where .opt cannot exist)

I'm rather ignorant on this topic. What are these platforms?

gares:
https://salsa.debian.org/ocaml-team/ocaml/blob/debian/sid/debian/native-archs
Hum, looks like almost everything is native these days, only the arm{el} and mips{el} are not.

>  I guess we should rather discuss it in a Coq-side issue, don't you think?

Yep, fine, although I'd first ask there and later fix here... that was the only point in my initial message

gares:
(that was written before the merge...)

maximedenes:
> Yep, fine, although I'd first ask there and later fix here... that was the only point in my initial message

Well, the thing is that I want to release VsCoq today or tomorrow, and as I wrote above, we need to support the current version of Coq. The discussion on the Coq side will apply to future versions, I imagine.

maximedenes:
> Hum, looks like almost everything is native these days, only the arm{el} and mips{el} are not.

I think VsCoq support for these platforms does not matter much, at least for now (note that arm64 and armhf are supported).

PR #69, opened 2019-10-03
[VsCoq1] Server is webpacked.

vlj:
This PR webpacks the server component ; I didn't find a way to solve the issue with webpacking the client though.

### DIFF
--- a/html_views/package.json
+++ b/html_views/package.json
@@ -16,10 +16,7 @@
   },
   "devDependencies": {
     "@types/jquery": "^3.3.30",
-    "shx": "^0.3.2",
-    "ts-loader": "^6.0.4",
-    "webpack": "^4.36.1",
-    "webpack-cli": "^3.3.6"
+    "shx": "^0.3.2"
   },
   "dependencies": {
     "jquery": "^3.4.1"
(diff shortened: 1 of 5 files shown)


Blaisorblade:
@maximedenes Did you forget to merge this or decide against it?

maximedenes:
Oh, I forgot, sorry about that! I'll test a bit and will see how we can combine with #89.

artagnon:
Can someone explain to me why this is a good idea? webpack adds extra bloat, along with a build-step. What does it buy us?

maximedenes:
https://code.visualstudio.com/api/working-with-extensions/bundling-extension

maximedenes:
IIRC, webpack made it simpler to load jQuery for the proof view. Note that utimately, the server should be replaced by a coq-side specific toplevel, so it does not matter so much.

artagnon:
Nice. It's good for performance then.

maximedenes:
> Nice. It's good for performance then.

Indeed. At least, that's what the recommendations say.

fakusb:
I think webpacking should be a good idea, it seems strange that currently, my local build consists of 8MB for this plugin.

Does anyone have an Idea on the current state of this PR? I've not set up TS/JS build environments at all.

thery:
This is superseded by #411

Issue #76, opened 2019-10-18
auto bracketting and comment


thery:
Maybe it has do with my config of vscode but if I want to write a comment.
I first type `(`. This gets automatically expanded to 
```()```
Then I type `*` and it gets automatically expanded to
```(**))```
It is a bit annoying I always have to remove this extra `)`

Issue #36, opened 2019-08-01
Feedback to empty Search


thery:
As Search goes to the output window it would be nice that there will be delimitators (specially when the result is void)

Search nat bool.

could output

"Search nat bool"  gives : 
====================

mult_n_Sm: forall n m : nat, n * m + n = n * S m

====================

pseudohuman92:
An option to clear the output before Search prints its result would be great too.

fakusb:
For the record: there is a button on the RHS of the "select Output" dropdown menu that clears the search manually, but an automatic option one seems like a good idea

thery:
@fakusb yes as a matter of fact my usual interaction is that  I do a `Search` then I realise that I can't figure out where the output start. So I do a clean and redo the `Search` :-(

PR #79, opened 2019-10-19
Remove comment delimitators from brackets list

GabrielaMafra:
Hey :octocat: 

VSCode matches `"(*"` with two brackets from the brackets list and then it is closed with both matches: `"*)"`and `")"`. Here I'm removing the comment deliminators from this list since its already declared as a `blockComment`.

This should fix https://github.com/coq-community/vscoq/issues/76


maximedenes:
> This should fix #36

I think you meant #76. Edited.

client/coq.configuration.json
```diff
@@ -7,7 +7,6 @@
 	"brackets": [
 		["{", "}"],
 		["[", "]"],
-		["(*", "*)"],
 		["(", ")"]
 	]
-}
\ No newline at end of file
+}
```
maximedenes:
```suggestion
,
	"autoClosingPairs": [
		{ "open": "{", "close": "}" },
		{ "open": "[", "close": "]" },
		{ "open": "(", "close": ")" },
		{ "open": "(*", "close": "*" }
	  ]
}
```

I think the change is sound, thanks! But I believe we can still keep the auto closing behavior for comments by adding this.

GabrielaMafra: Seems good! I wonder if matching works the same here (if it will match both `"("` and `"(*"`). I'm not sure how to build and test this. If you tell me how to or validate it yourself, that'd be great!

Thanks for the quick response!

maximedenes: I tried it and it seemed to work fine (it completes `(` with `)`, but if you then insert a `*`, it matches `(*` and inserts only the missing `*`.

GabrielaMafra: Nice! I've committed your suggestion :smile:

PR #89, opened 2019-12-20
[Modernize] Make it easier to hack on the extension from vscode

artagnon:
This series of commits makes it easier to work with vscoq from vscode. `F5` should directly build the extension launching it in "Extension Development Host", and `F8` should jump to the next error.

While at it, I have decided to update all dependent packages to the latest versions, and fix vscode to work with the new extension APIs. The build system has also been rewritten completely to use the (relatively new) TypeScript "composite" feature for the subprojects "client" and "server".


maximedenes:
This looks good, thanks, but I'd like to keep support for slightly older vscode versions (so that people can use the ones packaged by their distributions without issues). How hard would it be to stay compatible with say 1.39 ? It is not immediately clear to me where the new requirement comes from.

artagnon:
There is no hard dependency on 1.41; I just picked the latest. Fixed now: 1.39.0 is what we require.

maximedenes:
I will need a bit of time to review this, but it looks very nice.

artagnon:
No issues; take your time. Note that all my other PRs depend on this one.

maximedenes:
This PR is doing many things at the same time, it would be better to separate:
1. Upgrades of dependencies (and explain why they are necessary/useful)
1. Modernization of the build (making the project composite, flattening the layout)
1. Stylistic changes (and explain how they were generated and why they are an improvement)

I'm convinced that 2. is a clear improvement, so I'd like to have a PR with just that, in order to merge it quickly, if that's ok with you. If bumping dependencies is necessary for that, let's do it in a minimal way.

artagnon:
While I agree with you in principle, there are some practical considerations to keep in mind.

1. In the npm community, upgrading package dependencies is always considered a desirable change, since it often fixes security bugs and improves performance. The package-upgrading culture is very aggressive, and therefore, I thought bumping up the versions were a no-brainer.
2. 'composite' requires a newer version of typescript, and upgrading deps "in a minimal way" can be quite non-trivial. Considering (1), I'd say it's not worth the effort: the npm dependency-version-resolution algorithm isn't very robust in practice, and majority of problems are just fixed by upgrading to the latest versions of everything. On the off-chance that the latest version of one package is broken, there's always some form of help available on the interwebs concerning downgrading deps, since the userbase is so massive; the bug is usually fixed within a few days. Testing this PR would show that none of the upgrades are problematic.
3. Stylistic changes were necessitated by my "format-on-save" configuration with the `prettify` extension. I don't really care for stylistic changes as long as the style is consistent, and I assumed that nobody else would either, so I just separated out the stylistic changes after-the-fact into `0795128`. If you have strong objections against this, it's a little too late to do anything about it, since so much of my code depends on it, and rebasing would be ultra-painful.

Overall, I'd be tempted to merge this PR with minimal changes, after thorough testing, and not approach it with the same mindset of "conservative changes" present in other (more slow-moving) communities. I'm open to discussing any points that would improve the PR while not causing a rebase-headache.

Thanks.

artagnon:
Oh, a small piece of evidence that "upgrading everything to the latest versions is always a good thing" in the npm community: see the "Version Lens" vscode extension. I used it in this PR too.

maximedenes:
There was a bit of ambiguity in my previous message: upgrading packages is ok, what I would like to avoid is to bump the requirements too much (like the minimal supported version of vscode).

As a side note, I'm not so sure it is a good idea to blindly follow the common practices of the npm community. A patch changing thousands of lines of package dependencies (in lock files) can easily hide a security issue, for example. But again, this part is a minor comment.

Mixing the stylistic changes is a real problem to me, as it makes the patch essentially impossible to review. If it is too painful to remove it from this PR, I'll try to recreate a patch implementing the other, build-related, changes.

artagnon:
> There was a bit of ambiguity in my previous message: upgrading packages is ok, what I would like to avoid is to bump the requirements too much (like the minimal supported version of vscode).

In practice, vscode also auto-updates itself by default, and most people are running near-latest versions. We currently support 1.39 as per your request, but if you'd like, we can try shifting it back further.

> As a side note, I'm not so sure it is a good idea to blindly follow the common practices of the npm community. A patch changing thousands of lines of package dependencies (in lock files) can easily hide a security issue, for example. But again, this part is a minor comment.

In practice, security issues are reported by npm (and also by github itself) quite promptly.

> Mixing the stylistic changes is a real problem to me, as it makes the patch essentially impossible to review. If it is too painful to remove it from this PR, I'll try to recreate a patch implementing the other, build-related, changes.

Yes, that is a very valid concern, and I haven't mixed stylistic changes with code changes. The stylistic changes are all done in [0795128](https://github.com/coq-community/vscoq/pull/89/commits/0795128b3840a2c047deceffbb46cd5d5a1463bd), and the "NFC" stands for "No Functional Change", in case you missed that. You can check it for yourself in an interactive rebase: simply run `prettify --write` on the relevant files on the command-line and check the diff.

maximedenes:
Ok, let's take things one by one, then. The problem is that I don't like some of the stylistic changes, but I'd like to discuss them last, because the rest is much more important IMHO.

So let's start with updating dependencies and fixing the build. What command did you run to get the first commit (bumping the dependencies and formatting the `package.json` file)?

artagnon:
Okay. I used the `prettify` extension to format the files (you can run this on command-line), and the `Version Lens` extension to update all packages to their latest versions (in vscode, you simply install the extension and click to update button when viewing the various `package.json` files). As a side-job, I also fixed the SPDX license line, as I mentioned in the commit message.

artagnon:
> The problem is that I don't like some of the stylistic changes

[Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) has 4.6M installs, so I'd just roll with whatever style the community uses; it's not like I write enough TS to have even a weak opinion on style.

maximedenes:
Let's discuss stylistic concerns separately, for now I'm trying to have a working build with bumped dependencies.

artagnon:
> Edit: If you like, I can move the NFC patch to the top, so that the first commit excludes stylistic changes. This can be done easily, without a rebase-headache.

If you prefer, we can collaborate on [modernize-wip](https://github.com/artagnon/vscoq/commits/modernize-wip).

> Let's discuss stylistic concerns separately, for now I'm trying to have a working build with bumped dependencies.

You'd also need the various commits that come later in order to fix the build. Did you want every single commit in the PR to not break the build?

maximedenes:
> I can move the NFC patch to the top

Do you mean as the last or the first commit? What I would like is to get all the build-related improvements without discussing stylistic concerns for now.

> You'd also need the various commits that come later in order to fix the build.

Yes, I'm rebasing them.

> Did you want every single commit in the PR to not break the build?

Not necessarily, but I'd like to have one PR which addresses only one concern (bumping dependencies), which builds. Then, probably another one which changes the build layout. Finally, one which addresses stylistic concerns (I'll need to have a look at how Prettier can be configured, because there are a few things I don't like as is, both in the TS and in the HTML).

I'm almost done separating the upgrade part.

artagnon:
> Do you mean as the last or the first commit? What I would like is to get all the build-related improvements without discussing stylistic concerns for now.

First commit. I've done it in `modernize-wip`.
 
> > You'd also need the various commits that come later in order to fix the build.
> 
> Yes, I'm rebasing them.

> I'm almost done separating the upgrade part.

Great!

> > Did you want every single commit in the PR to not break the build?
> 
> Not necessarily, but I'd like to have one PR which addresses only one concern (bumping dependencies), which builds. Then, probably another one which changes the build layout. Finally, one which addresses stylistic concerns (I'll need to have a look at how Prettier can be configured, because there are a few things I don't like as is, both in the TS and in the HTML).

Okay. I had a different development style: I simply bumped up and broke everything, changed things around, and then fixed the build at the end.

maximedenes:
> Okay. I had a different development style: I simply bumped up and broke everything, changed things around, and then fixed the build at the end.

Sure, I understand, but it makes your PR look more like a development session than a semantic patch, TBH. In a collaborative development situation, it makes a difference, since it correlates the integration of otherwise orthogonal changes.

server/src/parsing/error-parsing.ts
```diff
@@ -154,79 +154,135 @@ Diff:
...
-    idx = indexOf(str,substr,pos);
+    pos = idx[0] + 1;
+    idx = indexOf(str, substr, pos);
   }
 }
 
-function costOfDifference(d: diff.IDiffResult[]) : number {
-  return d.reduce((v,x) => v + (x.added||x.removed ? x.value.length : 0), 0);
+function costOfDifference(d: diff.Change[]): number {
+  return d.length;
```
maximedenes: @artagnon is this the same thing as before?

artagnon: Yep, the API changed, and I think I found the exact equivalent: you've got to test it to be sure.

artagnon:
> Sure, I understand, but it makes your PR look more like a development session than a semantic patch, TBH. In a collaborative development situation, it makes a difference, since it correlates the integration of otherwise orthogonal changes.

Right, I've got to adjust to the level of effort I put into separating out patches for each community. I'll do a better job next time I write something for vscoq. To be honest, you've set the bar quite high for a vscode extension :)

Footnote: I've contributed to git in the distant past, and the level of discipline they demand is absolutely insane.

server/src/parsing/error-parsing.ts
```diff
@@ -242,13 +242,13 @@ export function parseError(txt: AnnotatedText): AnnotatedText {
...
         mid: [number, number];
-        diff: diff.IDiffResult[];
+        diff: diff.Change[];
         cost: number;
       }[] = [];
       for (let mid of indicesOf(str, df.mid, preLen, str.length - postLen)) {
         const x = str.substring(preLen, mid[0]);
         const y = str.substring(mid[1], str.length - postLen);
-        const xy = diff.diffWords(x, y);
+        const xy = diff.diffLines(x, y);
```
maximedenes: What about this change? Was it done on purpose?

artagnon: Yes. IIRC, there is no `diffWords` in the new API.

server/src/parsing/error-parsing.ts
maximedenes: This seems very strange: the previous cost model was depending on the length of inserted/removed string. How can that be reflected in the length of the array?

artagnon: TBH, I don't remember much of my reasoning; I wrote this a long time ago. However, there is an API reference [here](https://github.com/kpdecker/jsdiff) that you can use as a guide. I used `diffLines` instead of `diffWords`: that's the main difference I think.

server/src/parsing/error-parsing.ts
artagnon: Edit: I'm feeling unsure about this. Maybe it's better to use the new `diffWords` API to minimize the changes? See [ref](https://github.com/kpdecker/jsdiff).

client/src/HtmlLtacProf.ts
```diff
@@ -55,12 +55,14 @@ export class HtmlLtacProf {
...
-      httpServer.listen(0, "localhost", undefined, (e: any) => {
-        if (e) reject(e);
-        else resolve();
-      })
-    );
+    this.serverReady = new Promise<void>((resolve, reject) => {
+      httpServer.listen(0);
+      httpServer.on("error", (e: Error) => {
+        console.log(e);
+        reject(e);
```
maximedenes: This also looks incorrect: `resolve` will be called before any error event occurs, most likely, even if starting the server fails.

artagnon: I'll defer to your judgement on this, because I honestly didn't know what I was doing when I was editing this file: I don't know how Promises work.

maximedenes:
I think all the content of this PR was merged in separate ones, so closing.

PR #90, opened 2019-12-20
Indentation rules

artagnon:
Define some indentation rules.

~~This branch depends on the modernization of #89.~~

### DIFF
--- a/client/coq.configuration.json
+++ b/client/coq.configuration.json
@@ -1,19 +1,40 @@
 {
-	"comments": {
-		// symbols used for start and end a block comment. Remove this entry if your language does not support block comments
-		"blockComment": [ "(*", "*)" ]
-	},
-	// symbols used as brackets
-	"brackets": [
-		["{", "}"],
-		["[", "]"],
-		["(", ")"]
-	]
-,
-	"autoClosingPairs": [
-		{ "open": "{", "close": "}" },
-		{ "open": "[", "close": "]" },
-		{ "open": "(", "close": ")" },
-		{ "open": "(*", "close": "*" }
-	  ]
+  "comments": {
+    "blockComment": [
+      "(*",
+      "*)"
+    ]
+  },
+  "brackets": [
+    [
+      "{",
+      "}"
+    ],
+    [
+      "[",
+      "]"
+    ],
+    [
+      "(",
+      ")"
+    ]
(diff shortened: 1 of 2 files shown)


artagnon:
Rebased.

maximedenes:
@artagnon do you have an example of what this does, so that I try it out? Thanks!

artagnon:
Hi @maximedenes,

For the first rule, try typing "(*" and hitting <kbd>RET</kbd> -- you'll see "**" inserted on the next line. For the second, when typing out an Inductive, hit <kbd>RET</kbd> after a "|" line, and you'll see that it preserves indentation. For the third rule, type out "Definition foo :=" and hit <kbd>RET</kbd>, and you'll see that it increases indent. The patch isn't meant to be exhaustive -- just a first draft (or template) for others to extend.

maximedenes:
I'm not sure I fully understand the intent behind these rules:
- What is the expected shape of multiline comments? Currently, the rules generate:
```
(* one
** two
three
four *)
```
- What is the expected shape of inductive definitions? Currently:
```
Inductive foo :=
    | X
    | Y
```

I usually write :
```
Inductive foo :=
| X
| Y
```

artagnon:
Ah, yes. Poorly tested, because I rarely work on Real Things after locally launching the extension. Nevertheless, I'm guilty of not testing properly.

The comment thing: yeah, I should have included a rule for matching `**`; in retrospect, it's probably not a good idea, because `**` isn't unambiguous enough. Yes, for `Inductive`, your demonstration is indeed what is desired. It's intended for `Definition ... :=` and `Program Definition ... :=`, but I bundled in `Inductive` without thinking too much about it.

In any case, I was mainly writing a template for pushing in more indentation rules with little effort, and I'll get back to you after more testing.

maximedenes:
I'm not sure it is a good idea to add `**` at the start of each line, in particular I don't know if it is compatible with the coqdoc conventions.

If you like, you can just remove the rules for comments and `Inductive`, we can merge the remaining one (for `Definition` and the likes), and add other rules later.

artagnon:
Done.

artagnon:
@maximedenes Is there anything I can do to ease merging this?

artagnon:
Hey @maximedenes, could you merge this?

PR #91, opened 2019-12-20
[syntax] Allow unicode identifiers

artagnon:
~~Depends on #89.~~

### DIFF
--- a/client/syntaxes/coq.tmLanguage
+++ b/client/syntaxes/coq.tmLanguage
@@ -45,7 +45,7 @@
 
         <dict>
             <key>match</key>
-            <string>(Lemma|Theorem|Goal|Example)\s+([a-zA-Z][a-zA-Z0-9_']*)</string>
+            <string>(Lemma|Theorem|Goal|Example)\s+(\p{L}(\p{L}|[0-9_'])*)</string>
             <key>comment</key>
             <string>Theorem declarations</string>
             <key>captures</key>
@@ -65,7 +65,7 @@
 
         <dict>
             <key>match</key>
-            <string>(Program Definition|Program Fixpoint|Program CoFixpoint|Program Function|Function|CoFixpoint|Fixpoint|Definition|Let)\s+([a-zA-Z_][a-zA-Z0-9_']*)</string>
+            <string>(Program Definition|Program Fixpoint|Program CoFixpoint|Program Function|Function|CoFixpoint|Fixpoint|Definition|Let)\s+(\p{L}(\p{L}|[0-9_'])*)</string>
             <key>comment</key>
             <string>Definitions</string>
             <key>captures</key>
@@ -85,7 +85,7 @@
 
         <dict>
             <key>match</key>
-            <string>(CoInductive|Inductive)\s+([a-zA-Z][a-zA-Z0-9_']*)</string>
+            <string>(CoInductive|Inductive)\s+(\p{L}(\p{L}|[0-9_'])*)</string>
             <key>comment</key>
             <string>Inductive type declarations</string>
             <key>captures</key>
@@ -105,7 +105,7 @@
 
         <dict>
             <key>match</key>
-            <string>(Ltac)\s+([a-zA-Z][a-zA-Z0-9_']*)</string>
+            <string>(Ltac)\s+(\p{L}(\p{L}|[0-9_'])*)</string>
             <key>comment</key>
             <string>Ltac declarations</string>
             <key>captures</key>
@@ -170,7 +170,7 @@
 
         <dict>
             <key>match</key>
-            <string>\b(into|with|eqn|by|as|using)\b</string>
+            <string>\b(into|with|eqn|by|move|as|using)\b</string>
             <key>comment</key>
             <string>Common Ltac connectors</string>
             <key>name</key>
@@ -295,7 +295,7 @@
             <array>
                 <dict>
                     <key>include</key>
-                    <string>#block_comment</string>                    
+                    <string>#block_comment</string>
                 </dict>
                 <dict>
                     <key>include</key>


artagnon:
I just realized that this patch doesn't depend on the modernize branch, and have rebased onto master.

client/syntaxes/coq.tmLanguage
```diff
@@ -36,7 +36,7 @@
 
         <dict>
             <key>match</key>
-            <string>\b(Require|Import|Export|Local|Global|Open|Scope)\b</string>
+            <string>\b(Require|Import|Export|Local|Global|Open|Scope|Context)\b</string>
```
maximedenes: Why add `Context` here? TBH, I have no idea what this list was about in the first place.

client/syntaxes/coq.tmLanguage
```diff
@@ -65,7 +65,7 @@
 
         <dict>
             <key>match</key>
-            <string>(Program Definition|Program Fixpoint|Program CoFixpoint|Program Function|Function|CoFixpoint|Fixpoint|Definition|Let)\s+([a-zA-Z_][a-zA-Z0-9_']*)</string>
+            <string>(Program Definition|Program Fixpoint|Program CoFixpoint|Program Function|Function|CoFixpoint|Fixpoint|Definition|Let|Section|Arguments)\s+(([\p{L}-]|[0-9]|[_'])+)</string>
```
maximedenes: Why add `Section` and `Arguments` here? They are not definition-like commands.

client/syntaxes/coq.tmLanguage
```diff
@@ -125,7 +125,7 @@
 
         <dict>
             <key>match</key>
-            <string>\b(Hint|Constructors|Resolve|Rewrite|Proof|Ltac|Qed|Defined|Implicit|Set|Arguments|Tactic Notation|Notation|Section|Module Type|Module|End|Variables|Record|Check|Print|Eval)\b</string>
+            <string>(Hint|Constructors|Resolve|Rewrite|Proof|Ltac|Qed|Defined|Implicit|Set|Tactic Notation|Notation|Reserved Notation|Module Type|Module|End|Variables|Check|Print|Eval|Next Obligation)</string>
```
maximedenes: I see you change the list here too, but again, I'm not sure what's the idea behind this list. They are not keywords AFAIK, or at least not all of them.

client/syntaxes/coq.tmLanguage
```diff
@@ -45,7 +45,7 @@
 
         <dict>
             <key>match</key>
-            <string>(Lemma|Theorem|Goal|Example)\s+([a-zA-Z][a-zA-Z0-9_']*)</string>
+            <string>(Lemma|Theorem|Goal|Example)\s+(([\p{L}-]|[0-9]|[_'])+)</string>
```
maximedenes: Doesn't this allow identifiers beginning with numbers?

artagnon: Good catch; will fix shortly.

artagnon:
Better?

client/syntaxes/coq.tmLanguage
```diff
@@ -45,7 +45,7 @@
 
         <dict>
             <key>match</key>
-            <string>(Lemma|Theorem|Goal|Example)\s+([a-zA-Z][a-zA-Z0-9_']*)</string>
+            <string>(Lemma|Theorem|Goal|Example)\s+([\p{L}-]([\p{L}-]|[0-9]|[_'])+)</string>
```
maximedenes: I'm not very familiar with vscode's regexps, what does the first hyphen mean in the pattern?

artagnon: `\p{L}` matches `à`: I've just left out the end of the range, which would be the last unicode character. It's not a vscode-specific regex: you can look up any unicode regex [reference sheet](https://www.regular-expressions.info/refunicode.html).

maximedenes: But you want to match only letters, don't you?

maximedenes: More concretely, I'm wondering why `[\p{L}-]` rather than `\p{L}`.

artagnon: `θIO` is a valid identifier: getting it to highlight correctly was the main motivation behind this patch.

artagnon: Think of `[\p{L}-]` as the unicode equivalent of `[a-]` (or `[a-z]`). You don't want every identifier to begin with an `à`, right?

maximedenes: Doesn't `b` also match `\p{L}`?

maximedenes: Your documentation points to https://www.regular-expressions.info/unicode.html#category, which says: `\p{L} or \p{Letter}: any kind of letter from any language`

artagnon: Thanks for catching this: I read the documentation in a hurry; fixed now.

client/syntaxes/coq.tmLanguage
```diff
@@ -45,7 +45,7 @@
 
         <dict>
             <key>match</key>
-            <string>(Lemma|Theorem|Goal|Example)\s+([a-zA-Z][a-zA-Z0-9_']*)</string>
+            <string>(Lemma|Theorem|Goal|Example)\s+(\p{L}(\p{L}|[0-9]|[_'])+)</string>
```
maximedenes: And why do you exclude one-letter identifiers?

artagnon: Thanks again; fixed.

maximedenes: Also, you can keep `[0-9]` and `[_']` merged, as in the original code. (sorry for the multiple comments, I just noticed this)

artagnon:
I don't have write access, so could you merge this?

PR #92, opened 2019-12-20
Recommend latex-input

artagnon:
The latex-input is especially useful in Coq development. ~~Depends on #89.~~

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["yellpika.latex-input"]
+}


TheoWinterhalter:
What are you installing with it then?
Shouldn't it be more "recommended packages" instead? Input methods can also be handled by the system directly so it's not clear people would want to install it too (I do have it personally so I don't really mind).

artagnon:
Well, Proof General in emacs does all this, so don't you think it's okay for us? It's easy enough to revert if people complain, but I suspect the vscode people are going to be mildly surprised and pleased.

TheoWinterhalter:
Are you confusing with company-coq? PG doesn't come with LaTeX input methods if I remember correctly.
What else do you bring with this? There's too many lines you edited for me to look.

artagnon:
Sorry, I use Spacemacs with the Coq layer, and I assumed PG provided all these features.

Just have a look at [6cbbf57](https://github.com/coq-community/vscoq/pull/92/commits/6cbbf578ea8cf55039a6a1cc6410bf3cfb468926) -- the rest of the commits come from the modernize branch of #89.

TheoWinterhalter:
Ok, in that case I have another complaint (sorry… :(). I first installed the prettify module and had to uninstall it as it did not work really well in my experience. Replacing sometimes too much.
It's not as good as the company-coq prettifier, and I know a lot of people don't like prettifiers at all so again I'd be more keen on having an option here.

artagnon:
Okay, I can downgrade the dependency on `prettify-symbols-mode` to a recommendation. Is `latex-input` okay?

TheoWinterhalter:
I mean I'd be fine with it, but I'm not sure everyone is.
Also, having fake dependencies is a bit sad in itself isn't it?
Maybe some functionality to "install all recommended packages"?

Anyway I'm just one user so I'm not entitled to any decision, I'm just making it known that not everyone might be happy about this. (And some will probably be really glad about this as well.)

artagnon:
Why is it a "fake dependency"? It's just a batteries-included viewpoint, to get the user up-and-running as quickly as possible, with minimum fiddling. I would argue that you would want to manually *disable* the extensions if they don't work too well for you: by default, I think this is what most users would want.

Yes, vscode already prompts you to install recommended packages.

TheoWinterhalter:
I just mean that you would need those to install vscode, even though it's not strictly necessary.
And should one of those happen to be broken, it might make install fail.

Then why not just go with recommended packages because that's what they are?
Installing a bunch of packages just to disable them afterwards is a bit frustrating.

artagnon:
Fixed. Thanks for the productive discussion.

TheoWinterhalter:
I'm sorry for making you change everything.

artagnon:
Oh, it's not a problem: you're helping improve the PR.

Thinking about it a bit, `prettify-symbols-mode` is unmaintained, and making it a hard dependency is probably not a good idea.

artagnon:
Rebased.

artagnon:
Yeah, I agree with your comment about not including `prettify-symbols-mode`. I've been using it, and it's been giving me so many problems that I had to switch it off. I use the latex input for generating things like:

`Definition ∞category :=`

Spacemacs with the Coq layer allows one to type out symbols similarly, so I thought it would be a good thing for vscoq as well.

artagnon:
If the rationale sounds okay, have a look at my updated patch: we simply recommend latex-input.

artagnon:
A more conservative change now: is it okay to merge?

maximedenes:
@artagnon are you sure that extension is still maintained? No commit since Dec 2019, one open PR, no response since January...

I would prefer to let users pick their extensions than recommending some whose status is unclear to me.

artagnon:
Okay, let's close this.

PR #94, opened 2019-12-21
Convert grammar file to YAML

artagnon:
No functional changes. Depends on #89 ~~and #91~~.

#91 has been merged; now, it just depends on #89.


artagnon:
Should we pursue this patch? Apart from being a nicer format in general, it simplifies the regexes in the grammar file, so the double-escaping-backslash is not required. It adds an extra build-step though.

PR #95, opened 2019-12-22
Fix #24: Eliminate protocol.ts duplicate in server/client

### LINKED ISSUE #24: Share protocol definition between server and client
Today the file is duplicated (see `protocol.ts`), which makes the whole approach unsound.

artagnon:
~~Only the last two commits are new; it's based on the modernization of #89. Let me know when you merge #89, so I can rebase on top of it.~~

Rebased.


maximedenes:
Thanks for your PRs! I'll try to have a look soon, but I'm out of the office until Jan 2nd, so it may have to wait until then.

artagnon:
Should I work on removing the style changes from this? (the rebase is going to be painful)

PR #97, opened 2020-01-05
[HoTT] Acquire new setting for executing hoqidetop

artagnon:
To actually run with HoTT, depends on HoTT/HoTT#1176, which introduces the hoqidetop shim.

### DIFF
--- a/client/src/CoqDocument.ts
+++ b/client/src/CoqDocument.ts
@@ -288,6 +288,7 @@ export class CoqDocument implements vscode.Disposable {
 
   private async userSetCoqtopPath(global = false) {
     const current = vscode.workspace.getConfiguration("coqtop").get("binPath", "");
+    const coqtopExe = vscode.workspace.getConfiguration("coqtop").get("coqtopExe", "");
     const newPath = await vscode.window.showInputBox({ignoreFocusOut: true, value: current, validateInput: (v:string):string => {
       try {
         const statDir = fs.statSync(v);
@@ -298,16 +299,17 @@ export class CoqDocument implements vscode.Disposable {
       }
       let stat : fs.Stats|undefined = undefined;
       try {
-        stat = fs.statSync(path.join(v, 'coqtop'));
-      } catch(err) {}
-      if(!stat && os.platform()==='win32') {
-        try {
-          stat = fs.statSync(path.join(v, 'coqtop.exe'));
-        } catch(err) { }
+        stat = fs.statSync(path.join(v, coqtopExe));
+      } catch {
+        if (os.platform() === 'win32') {
+          try {
+            stat = fs.statSync(path.join(v, coqtopExe, '.exe'));
+          } catch {}
+        }
       }
-      if(!stat)
+      if (!stat)
         return "coqtop not found here"
-      if(!stat.isFile())
+      if (!stat.isFile())
         return "coqtop found here, but is not an executable file";
 
       return "";
@@ -316,7 +318,8 @@ export class CoqDocument implements vscode.Disposable {
       if(!newPath)
         return false;
       try {
-        return await fs.existsSync(path.join(newPath, 'coqtop')) || await fs.existsSync(path.join(newPath, 'coqtop.exe'))
+        return await fs.existsSync(path.join(newPath, coqtopExe)) ||
+          os.platform() === 'win32' && await fs.existsSync(path.join(newPath, coqtopExe, '.exe'))
       } catch(err) {
         return false;
       }
--- a/server/test/coqtop.8.7.ts
+++ b/server/test/coqtop.8.7.ts
@@ -41,6 +41,8 @@ describe("Coqtop 8.6", function() {
 
   const settings : proto.CoqTopSettings = {
     binPath: COQBIN_8_6,
+    coqtopExe: "coqtop",
+    coqidetopExe: "coqidetop.opt",
     args: [],
     startOn: "open-script",
   }
--- a/client/src/protocol.ts
+++ b/client/src/protocol.ts
@@ -32,6 +32,10 @@ export interface Settings {
 export interface CoqTopSettings {
   /** Path to coqc and coqtop binaries. @default `""` */
   binPath: string;
+  /** Name of coqtop binary. @default `"coqtop"` */
+  coqtopExe: string;
+  /** Name of coqidetop binary. @default `"coqidetop.opt"` */
+  coqidetopExe: string;
   /** A list of arguments to send to coqtop. @default `[]` */
   args: string[];
   /** When should an instance of coqtop be started for a Coq script */
--- a/package.json
+++ b/package.json
@@ -58,7 +58,17 @@
         "coqtop.binPath": {
           "type": "string",
           "default": "",
-          "description": "Path to coqc and coqtop binaries."
+          "description": "Path to coqtop and coqidetop binaries (alternatively, hoqtop and hoqidetop)."
+        },
+        "coqtop.coqtopExe": {
+          "type": "string",
+          "default": "coqtop",
+          "description": "Name of the coqtop binary (alternatively, hoqtop)."
+        },
+        "coqtop.coqidetopExe": {
+          "type": "string",
+          "default": "coqidetop.opt",
+          "description": "Name of the coqidetop binary (alternatively, hoqidetop)."
         },
         "coqtop.args": {
           "type": "array",
--- a/server/src/coqtop/CoqTop8.ts
+++ b/server/src/coqtop/CoqTop8.ts
@@ -199,11 +199,11 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
   }
 
   private get coqtopBin() {
-    return path.join(this.settings.binPath.trim(), 'coqtop');
+    return path.join(this.settings.binPath.trim(), this.settings.coqtopExe);
   }
 
   private get coqidetopBin() {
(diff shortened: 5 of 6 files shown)


artagnon:
Rebased, although `master` is still broken for me.

artagnon:
I'm not too pleased with this myself: why do we need to execute `coqtop` if `coqidetop` exists?

... but yeah, the refactoring is probably too much work. Let's stick to a short-term solution and change the instances (will update soon).

artagnon:
> More generally, isn't this an XY problem? Why does HoTT needs its own toplevel?

I assume that this has been resolved by your conversation on HoTT/HoTT#1176?

maximedenes:
> I assume that this has been resolved by your conversation on [HoTT/HoTT#1176](https://github.com/HoTT/HoTT/pull/1176)?

Yes, at least I think we should support a customized toplevel for now.

maximedenes:
> I'm not too pleased with this myself: why do we need to execute `coqtop` if `coqidetop` exists?

Indeed, there is some legacy code in VsCoq, we support older versions (like 8.7), where `coqidetop` does not exist, and we need to launch `coqtop -ideslave`. We can address this later, in another refactoring.

artagnon:
I think I've covered all the instances; do double-check though.

server/test/coqtop.8.7.ts
```diff
@@ -41,6 +41,8 @@ describe("Coqtop 8.6", function() {
 
   const settings : proto.CoqTopSettings = {
     binPath: COQBIN_8_6,
+    coqtopExe: "coqtop",
+    coqidetopExe: "coqidetop",
```
maximedenes: Why doesn't this match the documented default value?

artagnon:
Rebased. Please merge this and cut a patch release soon.

client/src/CoqDocument.ts
```diff
@@ -316,7 +317,7 @@ export class CoqDocument implements vscode.Disposable {
       if(!newPath)
         return false;
       try {
-        return await fs.existsSync(path.join(newPath, 'coqtop')) || await fs.existsSync(path.join(newPath, 'coqtop.exe'))
+        return await fs.existsSync(path.join(newPath, coqtopExe)) || await fs.existsSync(path.join(newPath, coqtopExe))
```
maximedenes: This line looks strange...

artagnon: Oops, fixed.

client/src/CoqDocument.ts
```diff
@@ -298,11 +299,11 @@ export class CoqDocument implements vscode.Disposable {
       }
       let stat : fs.Stats|undefined = undefined;
       try {
-        stat = fs.statSync(path.join(v, 'coqtop'));
+        stat = fs.statSync(path.join(v, coqtopExe));
       } catch(err) {}
       if(!stat && os.platform()==='win32') {
         try {
-          stat = fs.statSync(path.join(v, 'coqtop.exe'));
+          stat = fs.statSync(path.join(v, coqtopExe));
```
maximedenes: Aren't we trying a second time the same thing here?

artagnon: Fixed; sorry for the delay.

maximedenes:
I have a concern with this PR: doesn't it introduce a regression for Windows users? It seems to me that before the patch, VsCoq would guess the correct executable names for default Coq (`coqtop.exe` and the likes), and that after it will fail to do so (it will look only for `coqtop` and the likes).

Am I missing something?

artagnon:
I think it's resolved now, but have no way to test. Could you have a look?

client/src/CoqDocument.ts
```diff
@@ -298,16 +299,17 @@ export class CoqDocument implements vscode.Disposable {
...
-      } catch(err) {}
-      if(!stat && os.platform()==='win32') {
-        try {
-          stat = fs.statSync(path.join(v, 'coqtop.exe'));
-        } catch(err) { }
+        stat = fs.statSync(path.join(v, coqtopExe));
+      } catch {
+        if (os.platform() === 'win32') {
+          try {
+            stat = fs.statSync(path.join(v, 'coqtop.exe'));
```
Blaisorblade: Argh, seems my comments got lost. In short, this should probably use `coqtopExe + '.exe'` (tho that's arguable), and moving this fallback into the exception handler seems dangerous without sufficient testing (does stat only fail through exceptions, on all platforms? Is that actually guaranteed?) — it doesn't belong to the same PR at any rate.

client/src/CoqDocument.ts
```diff
@@ -316,7 +318,7 @@ export class CoqDocument implements vscode.Disposable {
       if(!newPath)
         return false;
       try {
-        return await fs.existsSync(path.join(newPath, 'coqtop')) || await fs.existsSync(path.join(newPath, 'coqtop.exe'))
+        return fs.existsSync(path.join(newPath, coqtopExe))
```
Blaisorblade: This doesn't check `coqtopExe + '.exe'`.

client/src/CoqDocument.ts
artagnon: See [statSync on node](https://nodejs.org/api/fs.html#fs_fs_statsync_path_options): the only failure mode is via an exception.

And no, it shouldn't be `coqtopExe` + '.exe'. We expect users to specify `coqtopExe` with a '.exe': I updated this patch mainly for backward compatibility; it's for Windows users who want VSCoq to work right out of the box, when 'coqtop.exe' is in `$PATH`.

client/src/CoqDocument.ts
artagnon: That is correct. If the user specifies a `newPath` via configuration, they'd be competent enough to specify a `coqtopExe` as well, no?

The VSCoq code is overcomplicated, and I intend to clean up this complexity along with my patches.

client/src/CoqDocument.ts
Blaisorblade: > We expect users to specify coqtopExe with a '.exe': I updated this patch mainly for backward compatibility; 

That’s not written in the setting description tho, is it?

artagnon: Good point. Fixed.

client/src/CoqDocument.ts
Blaisorblade: > The VSCoq code is overcomplicated, and I intend to clean up this complexity along with my patches.

Matter of style and not my call, but separating logically distinct changes helps when bisecting. And a simpler patch would require less testing to merge.
But if one is cleaning this up, why are there *two* variants of the logic? It also seems, from reading a bit more code, that `coqtop.exe` is tried when inputting a new path, but not here? If they’re inconsistent for good reason, shouldn’t that be explained in a comment?

> If the user specifies a newPath via configuration, they'd be competent enough to specify a coqtopExe as well, no?

Having dealt with students, I think “competent user” is too restrictive.
There’s a million valid reasons why a competent user would assume otherwise — most Windows programs just infer `.exe`, so writing `coqtop` is absolutely correct.
Worse, the default setting of coqtopExe works despite the `.exe` mismatch, so I expect that changing the binary folder (which is orthogonal) wouldn’t break this. The description is now a bit better, but it’s clearly incorrect enough I’d file a bug.

artagnon: > most Windows programs just infer .exe, so writing coqtop is absolutely correct

Oh, then we need the logic. Sigh.

artagnon:
Sorry for the delay, but I think the latest version should work?

Blaisorblade:
Since I commented earlier: I re-read again `userSetCoqtopPath`, and I'm not sure why the inserted path is checked two times in different ways and with different defaults — once we try `coqtopExe` and `'coqtop.exe'`, the other time we try `coqtopExe` and `coqtopExe + '.exe'`...
I'll send more comments...

Blaisorblade:
@artagnon @maximedenes Can you take a look at https://github.com/artagnon/vscoq/pull/1?

Those changes seem needed to address @maximedenes's comment: https://github.com/coq-community/vscoq/pull/97#issuecomment-592535825

I'm no TypeScript expert, so the actual code might be off — if so, please treat it as pseudocode. 

I still don't understand why the path is tested twice, but there could be a subtle reason, so both copies should be in sync.

As always: I'm not the reviewer, I'm just trying to help make the code more obviously correct.

maximedenes:
Sorry for the delay, having another look now. I even forgot why VsCoq needs `coqtop` (as opposed to `coqidetop`).

maximedenes:
> Sorry for the delay, having another look now. I even forgot why VsCoq needs `coqtop` (as opposed to `coqidetop`).

I remember now, it is for pre-8.9 Coq support.

artagnon:
Thanks to @Blaisorblade for the changes: merge-ready now?

maximedenes:
@artagnon the history is a bit messy now (it has merge commits), would you mind rebasing?

maximedenes:
For this particular instance, you can squash in one commit if that's easier.

artagnon:
Done. Please double-check before merging.

Blaisorblade:
Looking at https://github.com/coq-community/vscoq/compare/27d35bd30d4e4e8c7928170e04604a3fcd721ab5..7936c005d0693647d2f4966b2af73a3a4fba19ae, guess adding the binary package was not intended?

artagnon:
Oops, thanks a lot for catching that @Blaisorblade!

maximedenes:
Thanks! Hopefully we can make this whole logic simpler once we drop support for old Coq versions.

PR #101, opened 2020-01-12
Update deps again

artagnon:
In the effort to extract parts of #89.

PR #103, opened 2020-01-20
Update deps

maximedenes:
This is a rebased version of #101, excluding the formatting commit (I'd rather do a single pass on the whole codebase, but I need to tweak the settings first).

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -4,25 +4,19 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"@types/chalk": {
-			"version": "0.4.31",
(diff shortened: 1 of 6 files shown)


artagnon:
I'm sorry about the formatting changes -- I know that you don't like the defaults, but I'm left with the choice of turning off `format-on-save` (no `prettier` this time: I disabled it for the `vscoq` workspace) and suffering because either:

(1) I have no idea how to insert `\t` characters explicitly; some files are formatted with hard tabs.
(2) or having to endure the annoyance of hitting `<SPC>` multiple times to somehow match the quirky indentation of that particular file.

I don't care how the files are formatted, as long as I can edit files without either of those two annoyances, so please decide on some custom style and implement it.

PR #104, opened 2020-01-23
Flatten build

artagnon:
The third series extracted from #89. When this is merged, #89 can safely be deleted.


artagnon:
Note that we now require vscode 1.38 or higher. Otherwise, `vscode.d.ts` is too old, and the build fails.

tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es6",
+    "outDir": "out",
+    "lib": ["es6", "dom"],
+    "sourceMap": true,
+    // "forceConsistentCasingInFileNames": true,
+    // "noImplicitAny": true,
```
maximedenes: Why were these options changed? I don't think that's desired.

maximedenes:
This looks like a nice improvement, apart from compiler safety checks that were disabled, IIUC.

artagnon:
IIRC, the compiler safety checks got more aggressive with the typescript version bump, and many of them fail, which is why I disabled them (for now).

maximedenes:
Then we should either not bump or fix them, I'm afraid.

tsconfig.json
artagnon: My bad. The compiler options are different in `server`, `client`, and `html_views`, and I weakened checks by trying to unify them: fixed now.

client/tsconfig.json
```diff
@@ -1,23 +1,13 @@
...
-		"target": "es6",
-		"lib": [ "es6" ],
-		"module": "commonjs",
-		"outDir": "out",
-		"sourceMap": true,
-		"noLib": false,
-		"forceConsistentCasingInFileNames": true,
-		"noImplicitAny": true,
-		"noImplicitReturns": true,
-		"noFallthroughCasesInSwitch": true,
```
maximedenes: There are still a few flags that are gone, right? Like `noFallthroughCasesInSwitch`, `noUnusedLocals`.

maximedenes: Ah no sorry, I see they were moved to the toplevel file.

package.json
```diff
@@ -1,794 +1,829 @@
...
+  "author": {
+    "name": "Christian J. Bell",
+    "email": "siegebell@gmail.com",
+    "url": "http://people.csail.mit.edu/cj/"
+  },
+  "version": "0.3.0",
+  "publisher": "maximedenes",
+  "license": "MIT",
+  "engines": {
+    "vscode": "^1.38.0"
```
maximedenes: I changed this back to `1.23.0`, ran `make clean && make`, and it succeeded. What did I do wrong?

artagnon: Try:

```
$ git clean -dfx
$ npm i
$ npm run compile
```

IIRC, it downloads an outdated definition file and fails to build.

maximedenes: Oh, right, so it means there's a problem with the `clean` target. Not sure if introduced by this PR.

artagnon: I hadn't noticed the Makefile until now. Seems brittle.

maximedenes: > Yeah, the Makefile is junk; we should throw it out.

Why? What's wrong with it?

artagnon: The primary purpose of a Makefile is to incrementally build artifacts that are out-of-date. This Makefile has exactly one real dependency:
```
compile: node_modules
```

Obviously, this is incorrect, and `compile` should rebuild if anything in `client` or `server` changes. Besides, we now have incremental builds via `npm`: `tsc -b`. Makefiles are really ill-suited to js/ts projects, where npm is the build-tool of choice.

maximedenes: Fair enough, but let's address this in another PR. In particular, we need to add a call to `vsce` from `npm` before removing the Makefile.

Meanwhile, we can just adjust the directories to be erased. I'll push to this branch.

artagnon: No, `vsce package` is only useful to the extension publisher (namely, you): you shouldn't make contributors suffer slower build times for no benefit.

maximedenes: Yes, but there should be a build target for it, it does not have to be the default one.

artagnon: Why though? Just use `vsce publish patch` or similar directly -- why do you want a `vsce package`, and an even longer `npm run package`?

maximedenes:
Ok, with the 2 commits I added, this should be ready to merge. We can remove the Makefile in another PR.

artagnon:
Cool, thanks.

maximedenes:
Something is very strange: this break my VsCoq. Did you try to build and install the extension (i.e. the `.vsix` file)?

maximedenes:
Even when running the extension via F5, I can't get a proofview to be shown. What can be different in my setup?

artagnon:
The only thing I can think of is 1.30 versus 1.38?

artagnon:
Run `git clean -dfx`, `npm install`, and `npm run compile`. Make sure it
goes through the webpack step.

maximedenes:
Did you try with the `vsix` installation?

artagnon:
Give me a few minutes to try that. Maybe deps versus devDeps?

artagnon:
In any case, CI is broken. If it passes on Travis, it should be good to
merge. Could you add whatever package steps you do to CI? Also, we should
resurrect the test target so we have some basic sanity tests.

maximedenes:
> If it passes on Travis, it should be good to merge.

Certainly not, since we have no tests.

artagnon:
Yeah, it's very strange: the extension doesn't even activate.

artagnon:
Thanks for testing and catching this! I forgot to cherry-pick the patch changing paths in HtmlCoqView.

artagnon:
> > If it passes on Travis, it should be good to merge.
> 
> Certainly not, since we have no tests.

See `server/test` and `client/test`. My point is that we should wire them up to the CI, and write some more sanity checks, so that we can reasonably trust the patch when CI passes.

artagnon:
Damn it! The `vsix` still doesn't seem to work, and it's not a problem with devDeps versus deps: I've already checked that. This is frustrating.

artagnon:
Oof! It's finally done. I had to remove a stray line in `.vscodeignore`.

maximedenes:
Seems to work now, thanks!

PR #111, opened 2020-01-31
Fix _CoqProject parsing

maximedenes:
The `require` call did not work properly.


artagnon:
I think #104 broke this. Thanks for fixing.

PR #115, opened 2020-02-20
[VsCoq1] [.vscode] Do not format-on-save in this repository

artagnon:
Let's avoid churn for now. We can remove the setting later, when the
entire codebase has been formatted with a particular hand-picked style.


maximedenes:
Does vscode format on save by default, though?

artagnon:
I'm not sure, but for people with format-on-save toggled on (like me), this would be a good addition, yes?

Issue #123, opened 2020-03-13
Interrupt not working


jad-hamza:
Hello, the `Coq: Interrupt` command doesn't seem to be working (I don't see the highlight going away when I use it, and the state seems stuck there with the fan spinning), for instance on this example:

```coq
Ltac f := f.
Goal False.
  f.
```

Using this change it's working well: https://github.com/jad-hamza/vscoq/commit/5c05900898b00abde175e5b305f4263499331af7

Edit: using Linux

maximedenes:
Should be fixed by #127.

PR #127, opened 2020-04-12
Interrupt with SIGINT only

Blaisorblade:
Workaround #123 with a patch by the bug-reporter @jad-hamza. He didn't submit a PR, with [this argument on Gitter](https://gitter.im/coq-community/vscoq?at=5e6bef91034f6b7b24dd7c2d):*
> I don't think it's the right way to do it, I completely removed the XML part, I thought that could be discussed in the issue

Now a month passed, no progress has been made, and this patch improves things over master (I tried interrupting a failing `cbv`, and it works). So I suppose a proper discussion can be done either on this PR or after merging it. EDIT: in doubt, this PR might make things better enough to be worth merging anyway.

*Discussion starts [earlier](https://gitter.im/coq-community/vscoq?at=5e6b5b3517d3e74234782e8f).

### DIFF
--- a/server/src/coqtop/CoqTop8.ts
+++ b/server/src/coqtop/CoqTop8.ts
@@ -238,7 +238,7 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
   public /* override */ async coqInterrupt() : Promise<boolean> {
     if(!this.coqtopProc)
       return false;
-    else if(!super.coqInterrupt()) {
+    else {
       this.console.log('--------------------------------');
       this.console.log('Sending SIGINT');
       this.coqtopProc.kill("SIGINT");
--- a/server/src/coqtop/IdeSlave8.ts
+++ b/server/src/coqtop/IdeSlave8.ts
@@ -205,24 +205,7 @@ export class IdeSlave extends coqtop.IdeSlave {
 
   /** @returns true if an interrupt message was sent via the xml protocol */
   public async coqInterrupt() : Promise<boolean> {
-    if(!this.isConnected())
-      return false;
-    else if(this.useInterruptMessage) {
-      this.parser.once('response: value', (value:coqProto.ValueReturn) => {
-        this.console.log('interrupted');
-      });
-      this.console.log('interrupt');
-      
-      this.console.log('--------------------------------');
-      this.console.log('Call Interrupt()');
-      this.writeMain('<call val="Interrupt"><unit/></call>');
-      return true;
-    } else {
-      return false;
-      // this.console.log('--------------------------------');
-      // this.console.log('Sending SIGINT');
-      // this.coqtopProc.kill("SIGINT");
-    }
+    return false;
   }
 
   protected async checkState() : Promise<void> {


Blaisorblade:
FWIW, to test this PR I use `cbv` on a random goal from the Iris mode. I could try minimizing that an iris-using testcase, if needed, but I'm sure there are more self-contained ways; a looping typeclass search would be great.

jad-hamza:
Thank you for trying it and opening the PR, maybe someone knows the right way to adapt this. This is the example from #123 I used to test it:
```coq
Ltac f := f.
Goal True. f.
```

maximedenes:
What does CoqIDE do to interrupt Coq?

maximedenes:
> Now a month passed, no progress has been made

FTR, we are having a slight public health problem here since that date, which is impacting the amount of time I can spend on VsCoq. Expect degraded response times on both issues and PRs.

Blaisorblade:
Good question. Taking a look...

>> Now a month passed, no progress has been made
>
> FTR, we are having a slight public health problem here since that date, which is impacting the amount of time I can spend on VsCoq. Expect degraded response times on both issues and PRs.

@maximedenes I apologize, I phrased that very poorly; beyond the current emergency (we're also in lockdown here in the Netherlands), I didn't mean to complain, but my fault.

What I meant and failed to say was that, even tho who wrote the code thought it was a bad idea, this seemed worth submitting, also because a better solution wasn't coming.

maximedenes:
> I apologize, I phrased that very poorly; beyond the current emergency (we're also in lockdown here in the Netherlands), I didn't mean to complain, but my fault.

No pb, don't worry. I just wanted to make it explicit, so that users don't get the feeling that these bugs are not being addressed because we would not consider them as critical.

Blaisorblade:
Cool, thanks.

Back to what CoqIDE does, and AFAICT (and surely tons of people understand this code better and everything is with limited confidence):
- the XML code that this PR deletes uses an `"Interrupt"` command, which I cannot find used anywhere in CoqIDE's XML protocol. The string `Interrupt` appears, but it _seems_ only in the GUI (I could be wrong). I took a look at the various occurrences below (and also tried `grep -i`), but that's my best guess:
```
<from a coq checkout> $ git grep Interrupt ide/
ide/coq.mli:(** Interrupt the current computation of coqtop or the worker if coqtop it not running. *)
ide/coqide.ml:    ("Interrupt", "_Interrupt", `STOP, Nav.interrupt, "Interrupt computations", "Break");
ide/coqide_ui.ml:\n    <menuitem action='Interrupt' />\
ide/coqide_ui.ml:\n  <toolitem action='Interrupt' />\
ide/microPG.ml:  mkE _c "c" "Stop"    (Action("Navigation", "Interrupt"));
```

- CoqIde uses SIGINT in the case we consider, and @jad-hamza found some code in Coq handling it (https://gitter.im/coq-community/vscoq?at=5e6b8c59d86c595eb1fe4fcd).
 — In addition, CoqIde also has some support for sending a `StopWorker` message through the XML protocol — maybe to interrupt the _background_ proof workers?
https://github.com/coq/coq/blob/master/ide/coq.ml#L184
https://github.com/coq/coq/blob/master/ide/coq.ml#L516-L525

AFAICS, StopWorker takes a string argument referring to the proof worker, so it seems the message would be something like:
`<call val="StopWorker"> [encode string here somehow] </call>`

------------------
Would it make sense to merge this PR (maybe after more testing?) and create a feature request for "StopWorker"?

mattam82:
It would be great to be able to interrupt indeed :)

Blaisorblade:
From testing: even with this patch, modifying code while it is under processing does not send an interrupt to Coq, it just enters an unrecoverable state. That shouldn't stop merging this patch, but it'd be cool to fix that issue as well.

Trebor-Huang:
> From testing: even with this patch, modifying code while it is under processing does not send an interrupt to Coq, it just enters an unrecoverable state. That shouldn't stop merging this patch, but it'd be cool to fix that issue as well.

I would suggest a temporary fix by preventing the code to be modified when Coq is processing...? Since this avoids the hassle of having to restart the whole thing..

Blaisorblade:
> I would suggest a temporary fix by preventing the code to be modified when Coq is processing...?

> Since this avoids the hassle of having to restart the whole thing..

First to be clear: the biggest problem (and what you mentioned on Gitter) is that interruption just doesn't work, and that part is fixed by this patch.

On the rest, my 2 cents, for consideration of who can try additional fixes.
- I realize I described the leftover problem incorrectly — the problem I observed is when modifying code that Coq is currently looping over. It might include modifying code before the processing point (not sure), again only as long as Coq is looping.
Either case happens much less often.

- Most of the time, instead, modifying code in the processed region works fine, and it doesn't even interrupt Coq (or lose progress) if you modify 
whitespace/comments. I find that pretty cool and useful, and it'd be nice to preserve that.

PR #128, opened 2020-04-12
Bump version

Blaisorblade:
The last release did not perform a version bump; hence VsCode tries upgrading the build product using the Marketplace release.

I'm not a JS programmer and I'm not sure of the proper way to do it. I also upgraded `package-lock.json` even tho `make` didn't.

I suppose this (done properly) should be part of the release workflow.


Blaisorblade:
Maybe @artagnon knows better how to do this?

PR #130, opened 2020-04-30
Fix for large context: the conclusion of the first goal is at the bottom of the proof view

thery:


maximedenes:
Seems to work fine, thanks a lot! Sorry for the delay...

Issue #82, opened 2019-11-21
Automatically switching ProofView tab


aqjune:
Hello all,

When I switch to a different Coq file, ProofView tab isn't automatically switched to the corresponding one. Is there any option or any plan for supporting this?

VSCoq is a great plugin. Thank you for your maintenance!

Blaisorblade:
FWIW: after using #143 for a while, I think it's a regression. Sometimes overriding the choices becomes impossible. In particular, while stepping through a proof on the left pane, it becomes impossible to view a different file on the right pane.

With the right keycombo, switching by hand is easy enough; if that's a burden, VsCoq could make it easier.
https://github.com/coq-community/vscoq/issues/144#issuecomment-642551449

aqjune:
Hi,
Maybe we can have an option that turns it on/off? I also frequently open a different file on the right pane when skimming over definitions and writing functions rather than proving something.
I'd like to try the latest version of VSCoq (the one that #143 is applied). What is the good way to do it? My VSCode isn't showing its latest version. :/

Blaisorblade:
#143 was applied, but I find it an extremely strong regression, without a way to override it. Especially because the autofocus triggers without actually stepping through a proof.

Blaisorblade:
This is still a problem, and @thery might have asked about this in https://github.com/coq-community/vscoq/pull/327:
 
> Another question, when I have several ProofView, I'd like that the one that receives a new goal to be visible. What is the proper way to do this?

The present issue has some history on this: I think we fixed that a while ago, but reverted the fix (in part because of me)...  https://github.com/coq-community/vscoq/issues/144 links to https://github.com/coq-community/vscoq/pull/143 and https://github.com/coq-community/vscoq/pull/82.

These days I'd like at least a shortcut to open the matching proof view in one step...

PR #143, opened 2020-06-10
Auto-focus ProofView when its content updated

JimmyZJX:
This should satisfy most people regarding #82 :

When user navigate through any proof script, the proof status is updated, and thus the corresponding ProofView should focus automatically (or to call the `reveal` method).

Notice that I did not take care of the ProofView position for the moment; it should remain on its original view column. If that happens to be in the same column as the source script, the user should be aware and solve this problem by herself easily.

### DIFF
--- a/client/src/HtmlCoqView.ts
+++ b/client/src/HtmlCoqView.ts
@@ -160,8 +160,10 @@ export class HtmlCoqView implements view.CoqView {
   }
 
   private async sendMessage(message: ProofViewProtocol) {
-    if (this.panel !== null)
+    if (this.panel !== null) {
+      this.panel.reveal(this.panel.viewColumn, true);
       this.panel.webview.postMessage(message);
+    }
   }
 
   private async updateClient(state: proto.CommandResult) {


Blaisorblade:
For the record, this was reverted in https://github.com/coq-community/vscoq/pull/195.

thery:
@Blaisorblade I don´t  understand how do you see past goals when moving the cursor. Is this an option?

Blaisorblade:
@thery It a setting is involved, it might be `coq.autoRevealProofStateAtCursor`, which has `"default": false`:

> "markdownDescription": "If `true`, then auto-update the proof view with the cached state of the sentence the at the cursor. States are only cached when stepping through a proof, so the proof view will not be updated if e.g. the sentence was processed asynchronously and never had direct focus."

Blaisorblade:
- I do have `coq.autoRevealProofStateAtCursor` enabled and that seems to match what I wrote 2 years ago, but I don't think I realized the connection back then.
- Given how many like the behavior in this MR, we should probably have it at least as an option (even `default: true`). If we can skip tweak it so that `coq.autoRevealProofStateAtCursor` updates don't switch the view that's even better.

Blaisorblade:
(But both of us should probably move the discussion back to #82 rather than a merged MR).

Issue #135, opened 2020-05-19
using master branch


thery:
Hi,

I have installed vscoq as an extension so I have 0.3.1. How do I do to make vscode understand I want to use the master version?

thery:
@Blaisorblade  Thanks, worked like a charm. Created #145.

PR #145, opened 2020-06-11
doc local install

thery:
add a section in the README about local installl


thery:
the text comes from #135

maximedenes (approved):
Better than nothing. I guess we can improve the formatting later.

maximedenes:
@thery Thanks!

PR #149, opened 2020-07-16
Deleted outdated TODO.md

Blaisorblade:
I'm not sure, but it made a bad impression on a user; any current TODOs could be moved elsewhere.


TheoWinterhalter:
Shouldn't they be in the "Projects" tab maybe?
Why it would make a bad impression though?

Blaisorblade:
Let me rephrase: A user noticed this TODO was outdated when searching for the roadmap (a few days ago) so I created this branch, but forgot to actually create a PR till today.

But I agree users could file issues for some of those already — or maintainers could use whatever other tracking (including Projects if they so prefer).

TheoWinterhalter:
I see, I see.

maximedenes (approved):
I agree it is better to remove it until we publish something more up-to-date. Thanks!

PR #160, opened 2020-09-16
[Snippets] Remove "Solve All Obligations with ..."

artagnon:
Every time I try to `match`, ending with `with`, vscoq pulls out this
snippet. It's very annoying, and the value of this snippet is greatly
outweighed by its interference with a very common construct.


maximedenes (approved):
Sure, if this snippet is annoying, we can remove it for now.

PR #165, opened 2020-10-29
Improved query guessing (Search, Check,...):

fakusb:
Improved how vscode commands for Check, Search.. try to find the query when nothing is selected:

- If no selection exists, the current cursor are now searched for qualified identifiers:
So on cursor in `Init.Nat.add`, we match the whole thing instead of only `add`.

- If cursor is not in identifier, we are probably in a notation: We search for non-whitespace fraqment and wrap it into quotes as default query
So on cursor in `<=`, we produce the default query `"<="`, which can be used for example with About or Locate.

Initially, I also wanted to add Quotation marks around notations that are selected, but as it seems very hard to decide whether we selected something that is a notation or not, I did not change the behaviour when text is selected.

(One does probably not want to wrap `3+4` or `ident3` into quotes)


maximedenes (approved):
Wow! Isn't VsCoq doing better than CoqIDE for queries selection, with this patch? ;)

client/src/EditorAssist.ts
```diff
@@ -9,6 +8,8 @@ export function unload() {
   subscriptions = [];
 }
 
+export const regExpQualifiedCoqIdent = /((\p{L}|_\u00A0)(\p{L}|[0-9_\u00A0'])*)(\.((\p{L}|_\u00A0)(\p{L}|[0-9_\u00A0'])*))*/u;
```
Blaisorblade, line 11: @fakusb For some reason, Cmd-A on `_at_empSP_fwd.` gives `at_empSP_fwd not a defined object.` — ditto for `_at_empSP'_fwd` and `at_empSP'_fwd not a defined object.`

fakusb: Yes, that was an error, the () around `_\u00A0` was missing
fixed in c4878eabd806af64bb1933b1b723232f35670c40 , included in #175

Blaisorblade: ah, I need a newer version of your package, or @maximedenes should just make you maintainer and we could crowdsource the review

PR #166, opened 2020-10-30
Added options to disable italic/striked-out rendering of diffs

fakusb:
Two new options allow the user to disable the italic/striked out text style that is used for the ProofView-diff feature.


maximedenes:
@fakusb Do you think it would be disable these styles by default? I'm not a fan of it either.

fakusb:
I would disable them, for me the color is enough.

fakusb:
I disabled the options on default.

PR #167, opened 2020-10-30
Allow creating ProofViews for files that got closed before

fakusb:
This disables throwing an exception when a document gets registered for a file that already got registered before.

This fixes the behaviour that currently, after closing a file, the only file to get a proofView again is to restart VSCoq.

This maybe is a very bad hack; If yes, I want to open the PR and discuss what an actually fix would be.

Is this PR a Bad Idea because the language server in the background still has some resources and that would leak memory?
So, would a proper fix need to "remove" the CoqDocument from the language server when a file is closed?

### DIFF
--- a/client/src/CoqLanguageServer.ts
+++ b/client/src/CoqLanguageServer.ts
@@ -128,8 +128,8 @@ export class CoqLanguageServer implements vscode.Disposable {
   }
 
   public registerDocument(uri: string, doc: DocumentCallbacks) {
-    if(this.documentCallbacks.has(uri))
-      throw "Duplicate Coq document being registered.";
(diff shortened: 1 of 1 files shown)


Blaisorblade:
As a mere Coq user: I'm not against improvements, but any hack sounds great, as in "shut up and take my money" great.

> Is this PR a Bad Idea because the language server in the background still has some resources and that would leak memory?

Memory leaks in VsCoq are fixed by restarting VsCoq, so in the worst case, you'd need to restart VsCoq but less often. Unless you're leaking a GB per minute or sth.

fakusb:
I made a proper fix and open the PR soon

Issue #120, opened 2020-03-08
Bug: closing a file and re-opening it breaks interaction


bergwerf:
When I open a file interpret to the end, close it, and open again, then interaction is no longer working. When I press Alt + down it goes back to the file I was previously working on. If this is not reproducible I can provide an example. I use a `_CoqProject`.

mrhaandi:
This bug is quite troublesome because clicking a file once (and interpreting a part of it) opens it temporarily and then closes it as soon as another file is clicked.
After that one cannot reopen the closed file for interaction anymore.

bergwerf:
Double click to open a file permanently.

TheoWinterhalter:
Or "save" (<kbd>Ctrl</kbd> + <kbd>S</kbd>) in the tab (even without having made any change).

ljlin:
Same problem  on macOS 10.15.3 (19D76), VSCode 1.43.2, VSCoq 0.3.1 and Coq 8.9.0.
Double click or "save" does not solve this problem completely. It does work when I close a file for the first time. But the interaction stop working after I close for the sencond times and it goes back to the last file I previously closed.

bergwerf:
@ljlin That sounds exactly like what I described. The point of save/double click is to keep the file open. If you close it, interaction is broken.

Blaisorblade:
FWIW, keeping files open indefinitely is not a sustainable workaround — it just makes navigation impossible.

Cmd-Shift-P "Developer: Reload Window" reloads VSCode and VSCoq, fixing the VsCoq state — but then you must reprocess files in VsCoq.

fakusb:
FYI: PR #167 does allow you to reopen a file, but probably leaks memory.

fakusb:
And https://github.com/microsoft/vscode/issues/109749 asks for API for vscode that would allow to disable preview-mode once processing has started.

maximedenes:
Should be closed by https://github.com/coq-community/vscoq/issues/155.

PR #169, opened 2020-10-30
Files can now be reentered by coq after reopening

### LINKED ISSUE #155: Working with more than one Coq file does not work
Using VSCoq with just one Coq file open works reasonably well.
When I open a second Coq file it behaves fairly erratically. Some observable behavior:
- open file1.v + step down 1 step
- open file2.v + step down 1 step
- close file2.v
  * file1.v is gone (closed)
  * reopen file1.v shows that it has its state lost
  * step down in file1.v reopens file2.v
  * closing file2.v and reopen file1.v and step 1 step down again opens file2.v
  * As far as I can tell the only way out of this is to quit and reopen VSCode
- closing all open files before opening another one also doesn't help to avoid this.

In summary: I see now way to work with more than one Coq file except closing and reopening VSCode for each file.

fakusb:
Files can now be reentered by coq after reopening (while properly killing coqidetop)

Furthermore, files that had any interaction (goto cursor/end/position) that were only open as "preview" (single-click in explorer) files are promoted to "really open".

This improves on my former PR #167 and should close #155 and #120 .

### DIFF
--- a/client/src/CoqDocument.ts
+++ b/client/src/CoqDocument.ts
@@ -143,6 +143,8 @@ export class CoqDocument implements vscode.Disposable {
   }
 
   public dispose() {
+    this.quitCoq();
+    this.langServer.dispose();
     this.highlights.clearAll(this.allEditors());
     this.statusBar.dispose();
     if(this.view)
@@ -210,7 +212,7 @@ export class CoqDocument implements vscode.Disposable {
     this.statusBar.setStateReady();
   }
 
-  public async quitCoq(editor: TextEditor) {
+  public async quitCoq(editor?: TextEditor) {
     this.statusBar.setStateMessage('Killing CoqTop');
     try {
       await this.langServer.quitCoq();
@@ -219,7 +221,7 @@ export class CoqDocument implements vscode.Disposable {
     this.statusBar.setStateReady();
   }
 
-  public async resetCoq(editor: TextEditor) {
+  public async resetCoq(editor?: TextEditor) {
     this.statusBar.setStateMessage('Resetting Coq');
     try {
       await this.langServer.resetCoq();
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Sounds great! Follow-up question: can you check what happens if you close a file when coqidetop is crashed/stuck/...? Is the process still killed correctly? (Question inspired by all the times that Coq gets stuck and not interrupted correctly).

For a simple example:

```
Require Import Arith.

Lemma rewriteLoop (a b : nat) : a + b = 0.
Proof. rewrite !Nat.add_comm.
```

but you'll need a bigger file and async processing to get VsCoq in trouble.

fakusb:
> can you check what happens if you close a file when coqidetop is crashed/stuck/...? Is the process still killed correctly? (Question inspired by all the times that Coq gets stuck and not interrupted correctly).

Sadly, the proofworker becomes an orphan process and continues to diverge in the background. I think this is a bug of coqidetop, as coqidetop gets terminated correctly.

Blaisorblade:
That also sounds like an existing bug, I sometimes notice when those runaways when they eat too much memory.
But I thought somebody had fixed it by just sending a signal instead of also writing a message?

fakusb:
I think you mean #127 , but that talks about the vscoq interrupt command.
the part I talk about, "disposing" coqtop in case of e.g. closing th editor, already uses SIGKILL. and coqidetop also receives the sigkill, as it dies. but it seems that coqidetop does not properly kill its proofworker that worked on the diverging tactic.

Blaisorblade:
> the part I talk about, "disposing" coqtop in case of e.g. closing th editor, already uses SIGKILL. and coqidetop also receives the sigkill, as it dies. but it seems that coqidetop does not properly kill its proofworker that worked on the diverging tactic.

But coqidetop cannot handle SIGKILL to kill the proofworker, that’s not how SIGKILL works.

What about interrupting coqidetop (with the code in #127), then sending SIGTERM to tell coqidetop to clean up and exit? The only downside is if the clean up gets stuck, in which case SIGKILL is appropriate;
the POSIX semantics of sending SIGKILL to process P are basically to kill P with extreme prejudice, without letting P perform any cleanup. If any cleanup happens, that’s done by the OS.

In general, coqtop works well enough for Proof General, and coqidetop works well enough for CoqIDE — many of the VsCoq bugs are due to VsCoq sending the wrong commands (but probably not all). So if a solution is hard to find, checking what CoqIDE does might suggest solutions.

maximedenes (approved):
Looks like a nice improvement, thanks!

PR #140, opened 2020-06-05
Document manager

maximedenes:
This PR makes VsCoq use Coq's document manager (https://github.com/coq/coq/pull/12461) instead of adapting the CoqIDE protocol. This is very much work in progress.

PR #170, opened 2020-10-30
Fixed resizing issue

### LINKED ISSUE #151: ProofView occasionally switches to very narrow width
At some seemingly random points the proof view starts using only around 10 chars per line, even though the tab is much wider. 

Screen Shot 2020-07-30 at 11 30 31

fakusb:
Fixes #151 .

### DIFF
--- a/html_views/src/goals/goals.ts
+++ b/html_views/src/goals/goals.ts
@@ -60,7 +60,7 @@ const vscode = acquireVsCodeApi();
 
 function goalsLoad() {
 
-  window.onresize = throttleEventHandler(event => computePrintingWidth);
+  window.addEventListener("resize",throttleEventHandler(event => computePrintingWidth()));
   window.addEventListener("focus", onWindowGetFocus, true);
 
   window.addEventListener('message', event => {
--- a/html_views/src/goals/goals.ts
+++ b/html_views/src/goals/goals.ts
@@ -30,7 +30,9 @@ function computePrintingWidth() {
     const widthClient =stateView.clientWidth;
     const widthOneChar = ctx.measureText("O").width;
     let widthChars = Math.floor( widthClient / widthOneChar);
-    if (widthChars === Number.POSITIVE_INFINITY)
+    if (widthClient == 0)
+      widthChars = 80;
+    else if (widthChars === Number.POSITIVE_INFINITY)
       widthChars = 1;
     widthChars = Math.max(widthChars,5);
     $('#measureTest').text("<" + "-".repeat(widthChars-2) + ">");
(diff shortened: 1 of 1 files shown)


maximedenes:
Wow, this seems fairly straightforward! Pitty I didn't find the code to clean-up the code more systematically, I could have caught this problem. Of course, the long term fix is to format on the client side, but meanwhile this fix is very welcome.

Thanks, @fakusb ! This bug has been so annoying for VsCoq users that I'm considering doing a release even though I'm currently on vacations ;)

maximedenes (approved):

maximedenes:
I trust that you tested the fix, right? Not sure how easy it was to reproduce the bug.

fakusb:
I testes resizing, but reopening (#169 ) sometimes still leaves the width very small my beta-users reported. But one cab just resize te pnel to fix that.

I have to inverstigate this again, my guess is that the width gets computed before the html is "finished loading".

fakusb:
Ok, so I think I understood the reason the width is calculated very narrow on loading the webview sometimes, and I think I also have a solution. (The problem would also occur with #140 , as the problem is simply with how we can get the width of a webview/the editor containing the goal.)

The "onload" event is sometimes triggered before the webview rendered: https://github.com/microsoft/vscode/pull/85982#issuecomment-573640635
All elements still have width 0, which results in division by zero in `computePrintingWidth`.

The fix I made is to simply calculate&send the width once more after the webview got the first message. One could also call computePrintingWidth in goalsLoad after some delay using setTimeout, but I prefere this other solution. And my tester did not find a problem yet :)

Furthermore, I decided to use 80 characters width in all cases where the width could not be calculated, as this seemed a better default than 5 characters. One could also imagine a "minimal Width" and "maximal width" setting, but I don't think that is necessary.

Also, I had to substrat 27 pixels of width because of the goal passing, as otherwise, automatic resizing sometimes causes parts of the goal to go further than the right border.

I did not encounter any problems with this fix anymore.

fakusb:
Interestingly, the width after resizing does not take hypothesis names into account, so they could escape to the right.

This seems to be more than just a coding problem, but a conceptual question for which we might want the Language Server ;)

maximedenes:
> The problem would also occur with #140

In its current form yes, but in general we simply don't need to do this computation, Coq should just send HTML-like data. The rendering would be done on the client side (so in fact the computation would be done by the CSS engine in the webview). In particular, we wouldn't have to measure the width manually.

maximedenes (approved):
The patch looks good, and I find the new default (80 chars) more sensible indeed.

fakusb:
> In its current form yes, but in general we simply don't need to do this computation, Coq should just send HTML-like data. The rendering would be done on the client side (so in fact the computation would be done by the CSS engine in the webview). In particular, we wouldn't have to measure the width manually.

I always was under the impression that coq's notation system allows for some fancy annotations regarding printing (formatting boxes) that would be difficult to faithfully represent in css. But maybe that's just because I don;'t understand the annotations at all.

PR #171, opened 2020-11-01
Remove unneeded files and dependencies

fakusb:
I put unneeded files into the .vscodeignore and moved vsce (the extension packager) into the devDependencies, whish shrinks the resultin vsix file and speeds up packaging/installing/loading the extension.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -10,4 +10,4 @@ client/tsconfig.json
 server/**
 html_views/**
 out/**/*.js.map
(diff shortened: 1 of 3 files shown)


fakusb:
Together with the removal of the JS-dependency on unicode in #165 , the packaged extension now has 1/5 of the size and seems to install a bit faster.

maximedenes (approved):
Thanks, this should indeed help reducing the extension's size.

PR #172, opened 2020-11-03
Fix mangled diff

### LINKED ISSUE #164: ProofView uses distracting strikeouts, or mangles the term, when something failed to unify
When Coq fails to unify two terms, the vscoq ProofView gives me quite ugly output. It seems to be trying to "mark up" the left-hand term with some red backgrounds and strikeouts, as if to show me the diff between the two terms, but it doesn't help and just makes the term more-or-less unreadable. That much is true for the simplest cases, which I can cope with.

But in some cases, it makes real nonsense out of the term, as in the attached screenshot. It looks like a terminal that is randomly overwriting parts of previous lines. In this case, it's trying to saying something about a function called "Vector.at" but it's showing "r.ctor.at", among other problems. I'm on VSCoq 0.3.2.

Screen Shot 2020-10-24 at 2 41 08 PM

fakusb:
The problem was that diffs where computes without whitespaces, but applied to the whitespaced file.
Sadly, non-matching whitespaces are now highlighted in the diff as well. (While before they caused the diff to be totally messed up, so this is an improvement).

### DIFF
--- a/server/src/parsing/error-parsing.ts
+++ b/server/src/parsing/error-parsing.ts
@@ -200,7 +200,7 @@ function costOfDifference(d: diff.Change[]): number {
 }
 
 export function parseError(txt: AnnotatedText) : AnnotatedText {
-  const str = text.textToDisplayString(txt);
+  const str = text.textToString(txt);
   df: for(let df of diffMessages) {
     const preMatch = df.pre.exec(str);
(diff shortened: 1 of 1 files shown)


maximedenes (approved):
Not ideal, but seems better than before. Thanks!

PR #176, opened 2020-11-16
[VsCoq1] [.vscodeignore] Simplify ignore rules

artagnon:
We just need to move client/{snippets, syntaxes} to the root directory,
and then we can ignore all of client/.


artagnon:
@fakusb Could you review this? Thanks.

fakusb:
Does this make your workflow easier or is it "tidy up"-work? I was undr the impression of it beeing the later, and was afraid I might break things I don't test when including it in the most recent release.

artagnon:
Yes, it's more tidy-up work. Do test it first.

PR #177, opened 2020-11-16
[LICENSE] Collapse duplicates, modernize

artagnon:
ThirdPartyNotices does not apply anymore, so remove it.


fakusb:
I'm not sure we can remove the "third party" part:
https://github.com/coq-community/vscoq/blob/e1160bf151774b97eb3b05fb77431b385bf6d7bf/server/typings/pegjs.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pegjs/index.d.ts

artagnon:
Fixed.

PR #179, opened 2020-11-19
Added option for _CoqProject location

fakusb:
added option to give relative path of _CoqProject-files if not at root of vscode-workspace

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ The recommended way to install VsCoq is via the [Visual Studio Marketplace](http
 * `"coqtop.binPath": ""` -- specify the path to coqtop (e.g. "path/to/coq/bin/")
 * `"coqtop.args": []` -- an array of strings specifying additional command line arguments for coqtop
 * `"coqtop.loadCoqProject": true` -- set to `false` to ignore <span>_CoqProject</span>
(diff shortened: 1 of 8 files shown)


fakusb:
ups, this might breaks with actually making use of -R/-Q options, I made this a draft and investigate further.

artagnon:
Why not traverse up the directory tree until the project root directory, looking for `_CoqProject`? Otherwise, we have problems with making this new option point to a `_CoqProject` that's potentially in another project, no?

fakusb:
I don't think it is a problem if the user is able to shoot himself into the foot by setting the variable wrong, as long as the plugin allows for more possible workflows.

Looking up automatically as a default would still be an improvement, as it allows for example multi-project workspaces, I might look into it. Maybe as the default option if the coqProjectPath is empty.

fakusb:
I looked into it, and multiple coqProject-files in one workspace are more work than I'm willing to do.

Looking up the file for _CoqProject would also be nice, but the expedted time-spend to frustrated-user-time-saved ratio is not worth it for me.

So this PR won't alter anything more.
Still, thanks for the suggestion! If I wouldn't know that #140 will arrive at some time making all things done to the server obsolete, I would have thought more about it.

PR #180, opened 2020-11-20
Fixed task that debugs server

fakusb:
fixed the vscode-tasks allowing server debugging, and the mapping of ts-source-code to allow breakpoints in the server.


artagnon:
Yay! I have been wondering why I haven't been able to debug the server for a long time now :)

fakusb:
ste client=server task is still bigged, but the doing client first, and then doing "connect to server" allows for breakpoints in both parts at the same time.

Now I can finnaly check why the whole document gets reparsed whenever I add a new tactic to my large file

PR #181, opened 2020-11-20
Faster sentence parsing in huge files

### LINKED ISSUE #116: Performance issue on large files
The interface seems less reactive when working on a large file.
For example the file [CPoly_I.v](https://github.com/FlorianSteinberg/Cheby/blob/master/CPoly_I.v) has more than 3000 lines and inserting and executing
```
Goal True -> True -> True.
intros H1.
intros H2.
exact H1.
Qed.
```
at the beginning of the file takes ages.
It seems that at every execution, the ```outline``` window indicates ``` Loading document symbols for CPoly_I.v``` than after a while ```No symbol found for document CPoly_I.v```

fakusb:
VSCoq now checks if exactly one sentence (commands that would be send to coq, terminated by `. `) was removed or inserted and does not reparse the whole document, but reuses what was parsed before.

This behaviour was previously only there for changes within a single sentence.

This should fix #116.

I don't fully understand why the tests use "getDocumentEndOffset", but it works.

A fix that would work with larger changes (copy/paste) could maybe dynamically compute which old-sentence started at the position where we currently parse, and join dynamically, but I don't want to spend more time. Being able to edit coq files in the usual, add-or-remove-one-command way is a major improvement by itself, and suffices for me.

This should not be merged before end of next week, so i can properly test it more.

### DIFF
--- a/server/src/sentence-model/SentenceCollection.ts
+++ b/server/src/sentence-model/SentenceCollection.ts
@@ -342,6 +342,7 @@ export class SentenceCollection implements vscode.TextDocument {
     }
 
     try {
+      var oldSentenceCandidate : number = start;
       for(let idx = 0; /**/; ++idx) {
         const parseText = this.documentText.substring(currentOffset);
         const sent = parser.parseSentence(parseText);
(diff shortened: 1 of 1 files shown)


fakusb:
(I still have some console-prints in the code, they should not be too bad and make debugging a lot easier if it does not work.)

server/src/sentence-model/SentenceCollection.ts
```diff
@@ -353,10 +353,31 @@ export class SentenceCollection implements vscode.TextDocument {
...
+            console.log("NOTICE: some editing change lead us to reparse until EOF.")
+          }
           return {removed: removed, added: reparsed, endOfSentences: true};
-        } if(idx >= minCount && start+idx < this.sentences.length && currentOffset+sent.text.length === this.sentences[start+idx].getDocumentEndOffset() && sent.text === this.sentences[start+idx].getText()) {
+        }
+        
+        var fixByLocalGlueing : undefined | number  =  undefined;
+
+        if(idx >= minCount && start+idx < this.sentences.length
+            && currentOffset+sent.text.length === this.sentences[start+idx].getDocumentEndOffset()
```
fakusb, line 365: Does anyone understand why we use getDocumentEndOffset here?

fakusb: Why is the offset of the already parsed, old sentences comparable with our current position after parsing new things?

fakusb:
Ok, now even copy/pasting works, and this generalisation even de-duplicated a bit more code.

In some cases, link when intending a lot of lines at the same time and then deleting an area starting before the intended lines and ending within it, vscoq still parses the whole document. Still, this is an improvement. This might be related to some optimisation of not re-parsing anything at all for whitespace changes, but i don't care to much, unless I notice this behaviour in my future work on large files.

I;m pretty sure that my edits can not lead vscoq to loose sync with the real document. Only sometimes, more is parsed than needed, but less than before this fix.

fakusb:
One should probably also add that locations in sentences that shift due to insertions do not get updated, so error locations are off.

fakusb:
I merge this because it is a huge time saver on large files.

Error locations are off, but reloading the file fixes this, and as far as I see there is no time loss compared to the old behaviour if accurate error-locations are needed and reloading is performed.

PR #194, opened 2021-01-12
[snap] fallback on binaries provided by the snap package

### LINKED ISSUE #192: [wish] Fallback for coqidetop.opt
I'm preparing a [snap package for Coq](https://github.com/coq/platform/pull/49) which will contain `coqidetop.opt` but, due to snap "limitations", its name will be `coq-prover.coqidetop`.

Would it be possible to have the default executable name used by vscoq take into account this other name as well (as a fallback)?
The point is that I'd like it to work out of the box, and not have to tell users to configure this and that in vscoq.

gares:
Fix #192

### DIFF
--- a/server/src/coqtop/CoqTop8.ts
+++ b/server/src/coqtop/CoqTop8.ts
@@ -4,6 +4,7 @@ import * as net from 'net';
 import * as path from 'path';
 import * as vscode from 'vscode-languageserver';
 import * as semver from 'semver';
+import * as fs from 'fs';
 
 import {ChildProcess, spawn} from 'child_process';
 import {CoqTopSettings} from '../protocol';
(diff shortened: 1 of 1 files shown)


gares:
@fakusb would you mind handling this PR? (it's my first line of typescript, seems to work fine here).

gares:
I think this PR needs to be fixed, since the snap package takes precedence over the coqtop one may have in PATH, which is usually not what snaps do.

server/src/coqtop/CoqTop8.ts
```diff
@@ -198,12 +199,22 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
     this.console.log('coqtop-stdout:' + data);
   }
 
+  private fs_prefer_exists(current:string, alternative:string) {
+    if (fs.existsSync(current)) return current;
```
gares, line 203: I should look for something like `exists_or_is_in_PATH`...

Zimmi48:
Not needed anymore: more aliases have been added to the Snap package.

PR #195, opened 2021-01-29
Revert "Auto-focus ProofView when updated"

Blaisorblade:
This reverts commit bce1c76ca9dfb757679256b64111739e5b8ec952: it's a huge regression, as it prevents looking at another buffer on the right-hand pane when processing a proof on the left pane.


fakusb:
There is the possibility to add this behaviour back via an option in case someone needs it, but this revert seems more reasonable with regards to my day-by-day use as well.

PR #228, opened 2021-05-14
Remove incompatible brackets

qcfu-bu:
The removed brackets are incompatible with VSCode parsing. VSCode does not work well with brackets that share starting symbols or ending symbols, which breaks bracket colorization extensions that depends on parsing.

### DIFF
--- a/client/coq.configuration.json
+++ b/client/coq.configuration.json
@@ -17,30 +17,6 @@
     [
       "(",
       ")"
-    ],
-    [
-      "Proof.",
-      "Qed."
-    ],
-    [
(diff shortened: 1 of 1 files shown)


TheoWinterhalter:
Does that mean that these would no longer be foldable?

qcfu-bu:
They will still be foldable. Foldability is controlled by something else.

qcfu-bu:
One feature that I've noticed that has been lost in this PR is the auto-unindent of Qed, Admitted, etc. Closing this for now until I find a better solution.

Issue #236, opened 2021-06-08
"Quote notation" for deep pattern matching incorrectly displayed


YaZko:
Hello,

Consider the following snippet:
```
Goal forall (P : (nat * (nat * nat) -> Prop) -> Prop), P (fun '(x,(y,z)) => z = 0).
```

With PG, the goal gets displayed as expected as:
```
forall P : (nat * (nat * nat) -> Prop) -> Prop,
  P (fun '(_, (_, z)) => z = 0)
  ```
  
  While with `vscoq`, it gets displayed as the much less readable (especially in bigger goals naturally):
  ```
  ∀ P : (nat * (nat * nat) → Prop) → Prop,
  P (λ '(_, p), let '(_, z) := p in z = 0)
  ```

herbelin:
That's strange, I would not believe Vscoq is doing some smart printing itself which would cut the `(_, (_, z))` pattern in two, so I suspect the output is produced by Coq itself after vscoq sent to it some configuration data. Is the module `Ut8` loaded?

YaZko:
No `Utf8` loaded no, I get the same thing in a completely fresh file. 

If it is relevant, I have VSCoq v0.3.4, and Coq 8.13.2.

herbelin:
OK, I found it. The option `printingMatching` should be set to `true` in [`server/src/stm/STM.ts`](https://github.com/coq-community/vscoq/blob/a3025293e77c2b90c38486cb6fc7503b57c4b2bd/server/src/stm/STM.ts#L122). (The value was wrong at some time in coqide and this has apparently been transferred to vscoq.)

That would be nice if you propose a PR I think.

YaZko:
Ah great, thanks a lot!

I certainly can look into proposing a PR, it looks like a trivial one and that would force me to try compiling and testing vscoq locally and understanding a little bit how it's built... I'll look into this during the week.

herbelin:
This also means that you can manually fix the default with `F1`, then `Coq: Display Raw Matching Expressions`.

YaZko:
Oh indeed, even better, thanks!

I'll go through the exercise of doing a clean locally tested PR nonetheless, that'll kickstart me to try and provide fixes to my future complains rather than just open issues :)

herbelin:
Actually, this means that, like in CoqIDE (see coq/coq#14062), it might be needed to reverse the name of the option into something like `Display Nested Matching Expressions`.

PR #237, opened 2021-06-10
[VsCoq1] Changing default setting for displaying a nested match 

YaZko:
Hello,

This PR is meant to address #236 following @herbelin 's pointer.

It simply reverses the default behaviour in displaying nested pattern matches such as the following one:
```
Goal forall (P : (nat * (nat * nat) -> Prop) -> Prop), P (fun '(x,(y,z)) => z = 0).
```
Currently, the raw match, as two nested matches, is displayed by default while we would almost always like the deep pattern using the quote to be displayed.

The name of the option was also kinda backward: when set to true, the quote based display is used. The name hence also changes from `RawMatching*` to `NestedMatching*`, again following @herbelin's remark and a similar fix in coqide (https://github.com/coq/coq/pull/14062/commits/1fdc02ef7bf0b3d98eb2c69ab5cf8cd84f77b668).

Note: This is the first time I touch from afar or up-close the code of an ide, so even if the patch seems completely trivial, my apologies in advance if I've messed up! Please let me know if so :)


fakusb (commented):
Does this PR change package-lock.json on purpose?

Issue #257, opened 2021-10-09
Does not work over code-server


lijiaqigreat:
Hi,
   It seems that vscoq does not work over remote vscode using code-server. While doing the same thing on local vscode works.
Repro step:
Install coq using apt (version 8.11)
Install code-server using official install script (vscode version: 1.30)
Run code-server
Install vscoq in official store
Restart code-server
open a simple .v file
Run "Open proof view"
**vscode does nothing.**
Interpret some simple code.
**vscoq shows it's interpreting and hangs forever.**

fakusb:
Very basic question, as I have not used code-server: code-server and coq are installed and running on the same physical machine?

lijiaqigreat:
Correct. The host installs code-server, coq, and almost everything. The client uses vscode in browser without installing anything to the client OS.

fakusb:
Ok, as far as I see, this would require some rewriting, as for example we currently access the file system of the host in the extension using the node-fs-package. (for example, to look for the _CoqProject).

Blaisorblade:
I've learned a (relevant?) alternative from @gstew5: the "Dev Containers" extension can deal with VsCoq by installing it _inside_ the container:
 https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers

(I'm on M1 and run into https://github.com/microsoft/vscode-remote-release/issues/8169, but that issue lists workarounds).

PR #266, opened 2021-12-25
[VsCoq1] Webpack Client and Server

fakusb:
The client and server are now webpacked. To make this all work out, I had to split the single 'out' folder into three `out` folders, next to the already existing three `src` folders. Also, now a few more files are ignored when bundling the extension as `.vsix` file for publishing in the marketplaces (`.vscodeignore`). And I could remove a dependency on `copy-webpack-plugin`, which more than halved the number of npm-packages we depend on.

The debugging of the extension itself works as before, including breakpoints and sourcemaps.

The extension now loads ~20% faster (from a startup time of 100 ms to 80ms), and is smaller (312kB `.vsix` file instead of >5MB), and consists of a lot less files (The warning upon packaging the vsix file is now gone, so the nudging from microsoft to webpack worked for me).

And when we want to support vscode-server, webpacking will be needed (so this is towards #257 ).

This superseeds #69.


Makefile
```diff
@@ -1,10 +1,10 @@
-world: compile
+world: node_modules
 	node_modules/.bin/vsce package
 
 clean:
-	rm -rf node_modules out
+	rm -rf node_modules client/out server/out html_views/out
```
artagnon, line 5: This line can be replaced by a terse `git clean -dfx`, if the `.gitignore` is correct.

package.json
```diff
@@ -803,12 +803,17 @@
...
-    "watch": "tsc -b -w",
-    "test": "node ./node_modules/vscode/bin/test"
+    "vscode:prepublish": "npm run package",
+    "package": "webpack --mode production --config webpack.config.js",
+    "package:dev": "webpack --mode development --config webpack.config.js",
+    "compile": "webpack --mode none --config webpack.config.js",
+    "compile:client": "webpack --mode none --config webpack.client.config.js",
+    "compile:html_views": "webpack --mode none --config ./webpack.html_views.config.js",
+    "compile:server": "webpack --mode none --config ./webpack.server.config.js",
+    "watch": "webpack --watch --mode none --config webpack.config.js",
```
artagnon, line 815: I would drop "compile:server", "compile:client", and "compile:html_views", as webpack is smart enough to compile only what's necessary: a `webpack` invoked from the toplevel directory should do whatever is necessary, and it's unlikely that any developer would invoke any of these three build targets individually.

Further, `--config webpack.config.js` can be stripped, as webpack automatically looks for a `webpack.config.js` by default. Perhaps add `--devtool hidden-source-map` to "package", and drop the extraneous `--mode none`, as the common config specifies this mode.

Finally, I might be tempted to drop the "package:dev" target, as it's unlikely to be used directly.

package.json
```diff
@@ -803,12 +803,17 @@
...
+    "clean": "rimraf client/out && rimraf html_views/out && rimraf server/out"
```
artagnon, line 816: This is duplicating code from the Makefile and unnecessarily adding a dependency.

tsconfig.json
```diff
@@ -1,13 +1,19 @@
 {
+  "extends": "./tsconfig.base.json",
   "compilerOptions": {
     "module": "commonjs",
     "target": "es6",
-    "outDir": "out",
+    //"outDir": "out",
+    //"rootDir": "src",
```
artagnon, line 7: Please strip these commented out lines.

server/tsconfig.json
```diff
@@ -1,7 +1,13 @@
 {
-  "extends": "../tsconfig",
+  "extends": "../tsconfig.base.json",
   "compilerOptions": {
+    "module": "commonjs",
+    "target": "es6",
```
artagnon, line 5: Consider factoring out these two lines into `tsconfig.base.json`.

server/tsconfig.json
```diff
@@ -1,7 +1,13 @@
...
+    "outDir": "out",
+    "tsBuildInfoFile": "out/tsconfig.tsBuildInfoFile",
     "rootDir": "src",
+    "lib": ["es6", "dom"],
+    "sourceMap": true,
```
artagnon, line 10: Consider factoring out these two lines too.

server/tsconfig.json
```diff
@@ -1,7 +1,13 @@
 {
-  "extends": "../tsconfig",
+  "extends": "../tsconfig.base.json",
   "compilerOptions": {
+    "module": "commonjs",
+    "target": "es6",
+    "outDir": "out",
+    "tsBuildInfoFile": "out/tsconfig.tsBuildInfoFile",
```
artagnon, line 7: Very nice!

shared.webpack.config.js
```diff
@@ -0,0 +1,59 @@
...
+module.exports = 
+  function withDefaults(/**@type WebpackConfig*/extConfig) {
+
+    /** @type WebpackConfig */
+  const defaultConfig = {
+    mode: 'none', // this leaves the source code as close as possible to the original (when packaging we set this to 'production')
+		target: 'node', // extensions run in a node context
+		//node: {
+		//	__dirname: false // leave the __dirname-behaviour intact
+		//},
```
artagnon, line 19: Please remove this dead code.

artagnon (changes requested):
Nice overall. Should have minimal conflicts with #276.

thery:
@fakusb @Blaisorblade @huynhtrankhanh @4ever2. I'd like to include this to the next release.  This is far beyond my competence, :disappointed: . Any ideas how to rebase this?

4ever2:
@thery I rebased this and it seems to work okay after a few minor fixes. But still need to test it more.

https://github.com/4ever2/vscoq/tree/webpack

thery:
@4ever2  Thanks a lot. I think it is worth being merged. I will test it.

thery:
This is superseded by #411

PR #268, opened 2022-01-03
fixed leaking error message when only shelved goals remain

fakusb:


artagnon:
I think #274 is a little nicer.

html_views/src/goals/StateModel.ts
```diff
@@ -190,7 +190,7 @@ export class StateModel {
             ])
             var elmnt = document.getElementById("firstGoal");
             elmnt.scrollIntoView({block: "center", inline: "nearest"}); 
-          } else if (state.backgroundGoals.before.length > 0 || state.backgroundGoals.after.length > 0 ||  state.backgroundGoals.next != undefined ) {
+          } else if (state.backgroundGoals!= null && (state.backgroundGoals.before.length > 0 || state.backgroundGoals.after!= null && state.backgroundGoals.after.length > 0 ||  state.backgroundGoals.next != undefined )) {
```
Blaisorblade, line 193: @artagnon By "minus the null and undefined hard-coding" you mean this?
```suggestion
          } else if (state.backgroundGoals && (state.backgroundGoals.before.length > 0 || state.backgroundGoals.after && state.backgroundGoals.after.length > 0 || state.backgroundGoals.next)) {
```

artagnon: Yeah, thanks. In js, null and undefined are distinct, and this is somewhat unpleasant. In ts, the recommended practice is to never explicitly check for null or undefined.

artagnon: Further, the checks `state.backgroundGoals.before` and `state.backgoundGoals.after` are redundant: they have type `Goal[]`, so they can never be null or undefined.

artagnon: The type of `backgroundGoals` is incorrect in `protocol.ts`:

```ts
  backgroundGoals: UnfocusedGoalStack,
```

should be

```ts
  backgroundGoals?: UnfocusedGoalStack,
```

If we upgrade our version of typescript, and clean up the code a bit, we'd get errors from the compiler, and not runtime errors.

maximedenes (approved):

html_views/src/goals/StateModel.ts
maximedenes: @artagnon it would be nice indeed

Blaisorblade:
Could somebody please release a fix? A colleague's struggling again against this today.

PR #274, opened 2022-02-19
StateModel: don't .before on undefined (fixes #273)

### LINKED ISSUE #273: Fix for better error messages when admitting is broken
#184 is broken. For the following code:

```coq
Definition admitAdmitted: 1 = 0.
admit.
```

coqide prints:

```
No more goals, but there are some goals you gave up:

1 = 0

You need to go back and solve them.
```

but vscoq prints:

```
TypeError: Cannot read properties of null (reading 'before')
```

... and becomes completely unusable thereafter. One needs to restart vscode.

artagnon:
In an attempt to improve errors in the goal window, #184 was recently
mereged. Unfortunately, it is broken. The fix is permuting a few
lines.

### DIFF
--- a/html_views/src/goals/StateModel.ts
+++ b/html_views/src/goals/StateModel.ts
@@ -189,9 +189,11 @@ export class StateModel {
               , createFocusedGoals(state.goals)
             ])
             var elmnt = document.getElementById("firstGoal");
-            elmnt.scrollIntoView({block: "center", inline: "nearest"}); 
-          } else if (state.backgroundGoals.before.length > 0 || state.backgroundGoals.after.length > 0 ||  state.backgroundGoals.next != undefined ) {
-            this.setMessage("There are unfocused goals.");
+            elmnt.scrollIntoView({block: "center", inline: "nearest"});
+          } else if (state.backgroundGoals) {
+            if (state.backgroundGoals.before.length > 0 || state.backgroundGoals.after.length > 0 || state.backgroundGoals.next) {
+                this.setMessage("There are unfocused goals.");
+            }
           } else if (state.shelvedGoals.length > 0 ) {
             this.setMessage("There are shelved goals. Try using `Unshelve`.");
           } else if (state.abandonedGoals.length > 0 ) {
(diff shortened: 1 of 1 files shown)


artagnon:
The cases are actually independent: `state.shelvedGoals` would have zero length, and `state.backgroundGoals` would be undefined in the case of shelved goals. In the case of unsolved goals, `state.backgroundGoals` would not be undefined, and would have a `.before` and `.after` defined on it. My local testing on shelved and unsolved goals are both fine with this patch.

The question then becomes: is there a case in which there are shelved _and_ unsolved goals? My intuition says no. Consider `eapply` (without `unshelve`) followed by `admit`. We're admitting the shelved goals, not any unsolved goals.

Blaisorblade:
> The question then becomes: is there a case in which there are shelved and unsolved goals?

I agree with @fakusb and I think that (at least in user terms) all cases are possible together: I can case-split and focus on a goal, case-split further and admit some goals, and then use a tactic that shelves further goals.

Without larger refactorings, can you just guard the “crashing” tests by checking if backgroundGoals is undefined?

artagnon:
Okay, done with local testing.

Blaisorblade:
Looks reasonable, and I now realize the similarity with #268. Merging (and releasing) either fix would be nice.

Blaisorblade:
Renewing my request. @maximedenes @fakusb Do you think you could add @artagnon to the committers?

maximedenes:
I'm back in business, I can take this PR.

maximedenes:
Before reading the patch, I was expecting it to be what's actually in #268. Is this one still considered nicer? If yes, how so? Doesn't it potentially change behavior more than the other patch? E.g. if `backgroundGoals is defined`, but the conditions `state.backgroundGoals.before.length > 0 || state.backgroundGoals.after.length > 0 || state.backgroundGoals.next` are not satisfied, we no longer consider the other branches.

artagnon:
Feel free to drop this and merge #268, minus the null and undefined hard-coding. I opened this before I saw #268.

Issue #226, opened 2021-04-27
message when admitting


orilahav:
Running 

```
   Lemma a : 1=1.
   Proof.  admit.
```

in CoqIde gives the nice message:

```
   No more subgoals, but there are some goals you gave up:
   1 = 1
   You need to go back and solve them.
```

While vscoq gives a wrong (?) message:
```
There are unfocused goals.
```

fakusb:
Thanks, #184will fix this

rtetley:
Keeping this alive for VsCoq 2

PR #276, opened 2022-02-20
Housekeeping: yarn berry, consistent codebase formatting, all possible upgrades

artagnon:
Yarn is a great package manager; it's fast, modern, and overall, much better than npm. I've switched the package manager, done a bunch of package upgrades, and tested that everything still works on a very complex codebase. The only change for package maintainers is to use `yarn vsce` in place of `vsce`, when publishing. A couple of unfortunate things, to be tackled in future PRs:

1. We're not yet ready for the Yarn linker, ala Plug n' Play. We continue using the npm linker in this PR.
2. We couldn't upgrade vscode-langaugeserver, vscode-languageclient, and typescript. Quite a few modernization changes are required.

The final patch is a formatting patch, whence the entire codebase is formatted using `prettier`. This should significantly ease development. It is recommended to use the prettier plugin in vscode in all future development.


fakusb:
I have some refactoring in-progress regarding webpacking the whole extension (#226).

I'm a bit sceptical on changing npm to yarn; why change something that's not broken?

artagnon:
I think you meant #266. I had a look at it, and we can consider postponing the prettier patch until that lands.

While I agree that npm is not broken per-se, switching to yarn makes it easier for us to upgrade packages and maintain VSCoq. The dependency resolver in yarn is much better than the one in npm, and a simple `yarn upgrade-interactive` displays a ncurses window showing the safe and unsafe upgrades. In particular, the versions of typescript, vscode-languageserver and vscode-languageclient are very outdated in VSCoq, and we ought to upgrade them if the available documentation on VSCode is to be useful to us. The LSP protocol keeps evolving, and it's really quite irksome to find the documentation available on it to be newer than what we use.

Please consider switching to yarn, so I can attempt to upgrade the three aforementioned packages in a manner that is safe and pleasant; I will drop the formatting changes in view of #266.

PR #285, opened 2022-03-09
[VsCoq1] Add thread count option, translated to -async-proofs-j

Blaisorblade:
Sketch fix to coq-community/vscoq-legacy#48. This seems to work but:
- it's the quickest thing I could put together
- I don't know what I'm doing :-)
- I mostly just opened this code in VSCode, grepped for `coqtop.args`, and changed the code in what seemed the obvious way.

I chose to leave out `-async-proofs-j` by default, but I'd hope that `-async-proofs-j 1` is a safe no-op?

TODO:
- [x] revise help text (help and bikeshedding welcome)
- [ ] ~~it'd be nice to autodetect the thread count by running `nproc`! (More advanced). EDIT: looking at coq-community/vscoq-legacy#48, we'd need to at least look at free memory for a default :-).~~

### DIFF
--- a/server/src/CoqProject.ts
+++ b/server/src/CoqProject.ts
@@ -80,6 +80,9 @@ export class CoqProject {
 
     this.notReady();
     this.settingsCoqTopArgs = newSettings.coqtop.args;
+    if (newSettings.coqtop.threadCount) {
+      this.settingsCoqTopArgs = this.settingsCoqTopArgs.concat(["-async-proofs-j", newSettings.coqtop.threadCount.toString()]);
+    }
     this.currentSettings = newSettings;
 
     if(newSettings.coq.coqProjectRoot ){
--- a/package.json
+++ b/package.json
@@ -75,9 +75,12 @@
           "default": "coqidetop.opt",
           "description": "Name of the coqidetop binary (alternatively, hoqidetop)."
         },
+        "coqtop.threadCount": {
+          "type": ["null", "number"],
+          "markdownDescription": "Number of proof workers to enable."
+        },
         "coqtop.args": {
           "type": "array",
-          "default": [],
           "markdownDescription": "A list of arguments to send to coqtop. Use seperate elements instead of spaces to seperate each argument, especially when a flag expects another trailing argument, e.g. `[\"-I\",\"./bin\"]` instead of `[\"-I ./bin\"]`"
         },
         "coqtop.startOn": {
--- a/server/src/protocol.ts
+++ b/server/src/protocol.ts
@@ -37,6 +37,7 @@ export interface CoqTopSettings {
   coqtopExe: string;
   /** Name of coqidetop binary. @default `"coqidetop.opt"` */
   coqidetopExe: string;
+  threadCount?: number;
   /** A list of arguments to send to coqtop. @default `[]` */
   args: string[];
   /** When should an instance of coqtop be started for a Coq script */


gares:
maybe "proof workers" is more precise?
also, while we fo this, there is also the number of "tactic workers" for par: which should be settable.

gares:
it may also to be nice to tell the user that he can pass any options using the json interface (recalling him the key name).

Blaisorblade:
Thanks for the feedback. Is `-async-proofs-tac-j` documented? I can't find it in https://coq.inria.fr/refman/addendum/parallel-proof-processing.html, and I can't check if CoqIDE's dialog has docs (CoqIDE's config dialog crashes for me, there's a report); `ide` doesn't mention that option.

Are tactic workers separate, and if so do you get proof workers + tactic workers created?

MSoegtropIMC:
@Blaisorblade: as far as I know the number of threads is proof workers * tactic workers. Since use of the `par:` goal selector is rare, I guess most people won't benefit from it. In case we have a UI accessible option for it, it should be made clear that it only has an effect is `par:` is widely used. Again it might make sense to have this to promote the feature.

@gares : I think `par:` would be more useful if whatever number of workers is available is distributed over `par:` and proof jobs on a first come first serve basis. Otherwise the number of cases where using `par:` makes sense is fairly limited.

gares:
Totally agree but it is also harder to implement, since it is the proof worker process which spawns the tactic worker process.

MSoegtropIMC:
@gares: let's continue the discussion on Zulip - it is getting off topic.

server/src/CoqProject.ts
```diff
@@ -80,6 +80,9 @@ export class CoqProject {
 
     this.notReady();
     this.settingsCoqTopArgs = newSettings.coqtop.args;
+    if (newSettings.coqtop.threadCount != 0) {
+      this.settingsCoqTopArgs = this.settingsCoqTopArgs.concat(["-async-proofs-j", newSettings.coqtop.threadCount.toString()]);
+    }
```
artagnon, line 85: You can shorten this to `this.settingsCoqTopArgs = newSettings.coqtop.args.concat(["-async-proofs-j", newSettings.coqtop.threadCount.toString()])` -- no harm passing threadCount = 0.

... but I wonder if this is a good idea, considering that newSettings.coqTop.args might already contain a "-async-proofs-j", in the case where the user specified that argument in `settings.json` or `_CoqProject`.

Blaisorblade: I'd prefer to be conservative, I hope an extra `if` is maintainable.

Can this be easily shortened to something like `this.settingsCoqTopArgs ++= ["-async-proofs-j", newSettings.coqtop.threadCount.toString()]`?

artagnon: It's not about being conservative, or maintaining an extra if. The fact that there is a setting with a default value means that it should used. I can understand the check of the default value is undefined.

As far as I know, there is no such short syntax.

To address my other concern, probably change the setting to an override, without a default value, do an Array.search and Array.map to find the existing setting, and override it.

Blaisorblade: > The fact that there is a setting with a default value means that it should used. I can understand the check of the default value is undefined.

I'm not sure what you mean... Here I use "0" as default to mean "don't pass the option", I'm really encoding `option positive` and I'd be happy to have a better encoding, as long as it's well supported by the GUI.

> To address my other concern, probably change the setting to an override, without a default value, do an Array.search and Array.map to find the existing setting, and override it.

Also unsure what you mean, but:
- having this setting in `_CoqProject` would pass it to `coqc`, so like @gares I hope people don't do it.
- I doubt many people set this option for vscoq in 2 ways, trying to outsmart them seems unlikely to be safe. I'd prefer a warning.

MSoegtropIMC: I tend to agree with @Blaisorblade that a setting of 0 should mean there is no option. I am  actually not sure what happens if you say you want to have 0 threads - the behavior is definitely not documented, so it is not safe to assume it does no harm. But it is safe to assume that passing no option at all does not harm.

artagnon: All ts types are supported in the GUI. If you make it an optional int, the GUI will display a clear "unset", instead of a mysterious 0, which the user doesn't understand. There's nothing "safe" about assuming that 0 means unset: the user is going to read the man page of Coq to find out what this nthreads is, finding out that 0 is undocumented, and be confused about why VSCoq passes this option by default.

The design of options should be done in a way that doesn't confuse the user, and certainly doesn't allow them to shoot themselves in the foot. Yes, it is possible to set the option in more than one way, and the right way to handle this is to document one of the options as an override of another option, not to display a warning. Vscode is filled with overrides: for example, the default keybinding is this, this package overrides this keybinding, and finally, the user's bindings.json takes highest priority.

Blaisorblade: > All ts types are supported in the GUI. If you make it an optional int, the GUI will display a clear "unset", instead of a mysterious 0, which the user doesn't understand. 

Yesterday I didn't have time to try. This seems to work (based on https://code.visualstudio.com/api/references/contribution-points#contributes.configuration), but it took a few attempts:
```
        "coqtop.threadCount": {
          "type": ["null", "number"],
          "markdownDescription": "Number of proof workers to enable."
        },
```
I planned to document the `0` because I didn't know any better, so thanks for this. (I'll have to improve the description anyway).

FWIW, `coqtop.args` still refuses to display a proper editing widget :-|.

On the rest:

> The design of options should be done in a way that doesn't confuse the user, and certainly doesn't allow them to shoot themselves in the foot

Agreed, that's what we're trying to do.

> Yes, it is possible to set the option in more than one way, and the right way to handle this is to document one of the options as an override of another option, not to display a warning. Vscode is filled with overrides: for example, the default keybinding is this, this package overrides this keybinding, and finally, the user's bindings.json takes highest priority.

If there's a clear hierarchy, yes. For instance, for settings the workspace overrides the user config. But we were talking of conflicts between `_CoqProject`, this option and `coqtop.args` — possibly all set in the workspace. I could be convinced otherwise, but that's the definition of "trying to shoot yourself in the foot", probably unintentionally, and trying to guess what the user meant isn't helping. It also seems somewhat unlikely.

thery:

server/src/CoqProject.ts
Blaisorblade: Re-reading this thread, I think we can close it:
- I addressed one point by making `newSettings.coqtop.threadCount` optional.
- I'm against post-processing `coqtop.args` — VsCoq1 never does it, and any bugs in post-processing would block users. Bugs in user configs, _if_ they are bugs, can be fixed exclusively by user configs. Conflicts with `_CoqProject`... can be fixed by removing this option from `_CoqProject` since it doesn't belong there.

Blaisorblade:
I believe this is ready.

There was one point of contention, but I prefer to follow the user options accurately (as we always do) over trying to second-guess them in a potentially buggy way. Should you want to chime in, you can look into
https://github.com/coq-community/vscoq/pull/285#discussion_r1356034864.

gares:
I'm closing this PR since vscoq has now its own repository: https://github.com/coq-community/vscoq-legacy
Please reopen the PR there if you wish.

PR #300, opened 2022-07-14
Add hover provider

dlesbre:
Displays type in a tooltip when placing the mouse over an identifier.

This works by sending a `Check ${id_under_cursor}.` or `Locate ${notation_under_cursor}`, to the server and formatting the response to be shorter (just the type/notation def) using symbols.

Has such it has a few strong limitations:
- It only works for definitions that have already been interpreted.
- It doesn't work for local variables (introduced by `let v := ... in`, `fun v =>...` or `Definition foo v := ...`.

Coq doesn't allow you to redefine identifier in the same module, so most of the time the output of `Check ${name}.` will be the expected type. But there are a few conflicts with this method:
- A local variable can shadow a global definition. Here mouse-over on `foo` in `bar` shows `Type` and not `bool`
    ```coq
    Definition foo := Type.
    
    (* local conflict *)
    Definition bar (foo:bool) = if foo then 1 else 2.
    ```
- Definition in separate modules:
  ```coq
  Module A. Definition foo := true. End A.
  Module B. Definition foo := Type. End B.
  ```
  here interpreting to the second definition (not the end of B) and performing mouse-over on the first definition prints the second type.
- Similarly, values defined by an `Import` can shadow/be shadowed by local values.

Unfortunately I have no good idea on how to fix these issues. I do find the feature in it's current state is still fairly useful and if you have any suggestions to fix these issues I'd be happy to hear them.

Also of not, the hover provider is not disabled inside comments, so it will perform useless queries there. I don't think that's much of a problem though.

**Disclaimer:** I'm new to typescript and don't know much about optimal/efficient TS code or recommended coding styles.


gares:
Hello, thanks for this PR. I let @fakusb judge its integration in the main branch. I take the occasion to mention the [Roadmap](https://github.com/coq-community/vscoq/wiki/VsCoq-1.0-Roadmap) for what I hope will be the future of vscoq.

client/src/HoverProvider.ts
```diff
@@ -0,0 +1,179 @@
...
+  type = type.replace(/^ {7}/gm, ""); // remove indent
+
+  if (type === "") return;
+  type = compactify(type);
+
+  let hover = [{ language: "coq", value: type }];
+  // if (where)
+  //  hover.push({language:"coq", value:where});
+  return new vscode.Hover(hover);
+}
```
Blaisorblade, line 60: @dlesbre How hard w you replace `Check function.` by `About function.`? If `function` takes implicit arguments, `Check` can be somewhat unhelpful:

image

```coq
About set_map.
(*
set_map :
∀ {A C : Type},
  Elements A C
  → ∀ {B D : Type}, Singleton B D → Empty D → Union D → (A → B) → C → D

set_map is not universe polymorphic
Arguments set_map {A C}%type_scope {H} {B D}%type_scope 
  {H0 H1 H2} f%function_scope X
set_map is transparent
Expands to: Constant stdpp.fin_sets.set_map
*)
Check set_map.
(*
set_map
	 : (?A → ?B) → ?C → ?D
where
?A : [ |- Type]
?C : [ |- Type]
?H : [ |- Elements ?A ?C]
?B : [ |- Type]
?D : [ |- Type]
?H0 : [ |- Singleton ?B ?D]
?H1 : [ |- Empty ?D]
?H2 : [ |- Union ?D]
*)
```

Blaisorblade:
I'm testing this locally, and it seems indeed useful enough to merge — if we were worried about regressions, we could probably add another option flag, but for now I don't see a strong reason — despite comment about `Check` vs `About`.

> Unfortunately I have no good idea on how to fix these issues. I do find the feature in it's current state is still fairly useful and if you have any suggestions to fix these issues I'd be happy to hear them.

Those issues can be fixed when Coq itself will provide support for such editor services.

client/src/HoverProvider.ts
dlesbre: Replacing one with the other is fairly simple. I used Check more or less as a default since it's easy to parse if you just want the quick and short type.

I was concerned about hovers becoming too long (which is also why I cut and format the output of Check to keep it minimal).

I'll try using 'About' in a few hours

TheoWinterhalter: Maybe it could be a user setting to pick which? (Unless it's too much work of course.)

dlesbre: I added 'About' and it seems to work fine. I made the choice between Check and About a user setting that default to "about".

The hover using about also provides additional information (namely, the "Arguments" line which shows the scope and implicit status of arguments, and the "Expands to: " line)

thery:
Look good to me. If there is no reaction, I will merge it tomorrow.

Issue #134, opened 2020-05-17
Support dune builds


Blaisorblade:
When using dune, currently VsCoq won't be able to find and load built `.vo` files (ditto ProofGeneral/CoqIDE).
This is a known problem, but I'm filing this issue as documentation and to post the current workaround.
I'm using an issue as documentation on purpose, since this is still pretty hacky.

As confirmed by @ejgallego, the workaround is to modify _CoqProject as follows:

```diff
--Q theories D
+-Q _build/default/theories D
```

Beware this breaks tasks that use `coq_makefile`, but I expect on *nix one can try processing `_CoqProject` with `sed -e 's!_build/default/!!` before passing it to coq_makefile.

gares:
Thanks for posting. I guess the plan is to invoke coq*top via dune, since only dune knows the actual profile (which may not be `default`). In the meanwhile this is better than nothing ;-)

Blaisorblade:
Haven’t heard anything, and that MR seems about jump-to-definition only, which is unsupported altogether in vscoq.

fakusb:
Probably one would just need to detect if the current project is a dune-project, and prepend something to the command line arguments of coqtop that tell where to look for the libraries. And one probably want to have an option in vscode to force either of he two ways instead of the auto-detection.

I have never used dune. Can someone link a small project with the default setup for a dune-coq-project?

k4rtik:
@fakusb thanks for taking a look, here's a template project: https://github.com/palmskog/coq-program-verification-template

Especially notice the caveat at https://github.com/palmskog/coq-program-verification-template#caveats We are looking to improve this situation from VSCoq perspective, so one can imagine being able to work on this project with VSCoq when the` `Makefile*` and `_CoqProject` are deleted.

Currently, a suboptimal hack mentioned above in the OP is required.

gares:
See also #361

Zeta611:
Is there an update on this? The workaround
```
-Q _build/default/theories D
```
still works, but it took quite some time to find this issue page for the workaround. Or can we document this somewhere to make it more discoverable?

gares:
This is still TODO.
The correct solution would be to depend on dune-rpc and use this API:
https://github.com/ocaml/dune/blob/6152ab8ec5d7204d792be7b094063c032bfdd3cc/otherlibs/dune-rpc/dune_rpc.mli#L334

PR #319, opened 2022-12-12
[VsCoq1] Progress on #134: Use dune coq top (when enabled by a setting)

### LINKED ISSUE #278: vscoq passes incorrect `-topfile` argument
vscoq uses `-topfile file:///path` or `-topfile untitled:/URL`,  but Coq demands a path (relative or absolute, doesn't matter); that path is compared against the `-Q`/`-R` options to compute the correct name of  the module being created.

For instance, assume `_CoqProject` consists of `-Q theories D`, and `test_topfile.v` is the following:
```coq
Definition hello := Prop.
About hello.
Check D.test_topfile.hello.
```
This code (based on the testcase in https://github.com/ProofGeneral/PG/issues/407#issuecomment-447289254) compiles fine and works in `coqtop` with the correct options, but VSCoq fails to step through it.

Example incorrect command lines that VSCoq uses:
```
exec: /Users/pgiarrusso1/.opam/__coq-platform.2021.11.0~8.14~2021.11/bin/coqidetop.opt -main-channel 127.0.0.1:58104:58105 -control-channel 127.0.0.1:58106:58107 -async-proofs on -Q theories D -w -notation-overridden -w -redundant-canonical-projection -w -convert_concl_no_check -w -undeclared-scope -w -ambiguous-paths -w -deprecated-hint-rewrite-without-locality -w -deprecated-instance-without-locality -w -deprecated-typeclasses-transparency-without-locality -topfile untitled:/Users/pgiarrusso1/git/gDOT/dot-iris-make/theories/test_topfile.v
```

```
exec: /Users/pgiarrusso1/.opam/__coq-platform.2021.11.0~8.14~2021.11/bin/coqidetop.opt -main-channel 127.0.0.1:58158:58155 -control-channel 127.0.0.1:58156:58157 -async-proofs on -Q theories D -w -notation-overridden -w -redundant-canonical-projection -w -convert_concl_no_check -w -undeclared-scope -w -ambiguous-paths -w -deprecated-hint-rewrite-without-locality -w -deprecated-instance-without-locality -w -deprecated-typeclasses-transparency-without-locality -topfile file:///Users/pgiarrusso1/git/gDOT/dot-iris-make/theories/test_topfile.v
```

Blaisorblade:
Address #134 for VsCoq1. The first commit fixes #278.

This is a V0 prototype with some limitations, but it appears somewhat suitable for use.
- There's no auto-detection of `dune` projects, but an explicit setting. Since it can be set per workspace, since most dune projects _also_ have a `_CoqProject` file, and since `dune coq top` is probably somewhat experimental, I consider this sufficient for now.
- If the build fails, this does not redirect build outputs/errors to some proper log window (beyond "Coq Language Server" logs).
- User experience isn't great when dune decides to build code: a standard progress bar appears, but one must view "Coq Language Server" to understand what is happening.

What I tested was my integration branch — https://github.com/Blaisorblade/vscoq/tree/thread-count-option-misc.

Some unexpected messages appeared, but they seem due to the options added by dune coq top.
```
coqtop-stderr: Warning: Native compiler is disabled, -native-compiler ondemand option
ignored. [native-compiler-disabled,native-compiler]
Warning: The native-compiler option is deprecated. To compile native files
ahead of time, use the coqnative binary instead.
[deprecated-native-compiler-option,deprecated]

coqtop-stdout:Skipping rcfile loading.
```

The last one suggests that `.coqrc` is not being loaded, but it appears untrue — I used
```
Test Nested Proofs Allowed.
Test Ltac Backtrace.
```
to confirm my  `.coqrc` settings were active.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
         "coqtop.binPath": {
           "type": "string",
           "default": "",
-          "description": "Path to coqtop and coqidetop binaries (alternatively, hoqtop and hoqidetop)."
+          "description": "Path to folder with coqtop and coqidetop binaries (alternatively, hoqtop and hoqidetop)."
         },
         "coqtop.coqtopExe": {
           "type": "string",
--- a/package.json
+++ b/package.json
@@ -94,6 +94,15 @@
           "default": "open-script",
           "markdownDescription": "When to start an instance of coqtop: when a Coq script is opened (`open-script`) or else when the user begins interaction (`interaction`; default)."
         },
+        "coqtop.projectSettingsType": {
+          "type": "string",
+          "enum": [
+            "_CoqProject",
+            "Dune"
+          ],
+          "default": "_CoqProject",
+          "description": "Load project settings from _CoqProject, or use dune coq top."
+        },
         "coqtop.useDune": {
           "type": "boolean",
           "markdownDescription": "(Experimental) Use `dune coq top` to run `coqidetop`."
--- a/package.json
+++ b/package.json
@@ -60,6 +60,11 @@
       "type": "object",
       "title": "Coq configuration",
       "properties": {
+        "coqtop.dunePath": {
+          "type": "string",
+          "default": "dune",
+          "markdownDescription": "Full path to dune binary (`dune` included)."
+        },
         "coqtop.binPath": {
           "type": "string",
           "default": "",
@@ -88,6 +93,10 @@
           ],
           "default": "open-script",
           "markdownDescription": "When to start an instance of coqtop: when a Coq script is opened (`open-script`) or else when the user begins interaction (`interaction`; default)."
+        },
+        "coqtop.useDune": {
+          "type": "boolean",
+          "markdownDescription": "(Experimental) Use `dune coq top` to run `coqidetop`."
         },
         "coq.loadCoqProject": {
           "type": "boolean",
--- a/server/src/CoqProject.ts
+++ b/server/src/CoqProject.ts
@@ -89,6 +89,8 @@ export class CoqProject {
     if(newSettings.coq.loadCoqProject ) {
       this.watchCoqProject();
       await this.loadCoqProject();
+    } else {
+      this.currentSettings.coqtop.args = this.settingsCoqTopArgs;
     }
 
     if(newSettings.prettifySymbolsMode && newSettings.prettifySymbolsMode.substitutions) {
--- a/server/src/CoqProject.ts
+++ b/server/src/CoqProject.ts
@@ -86,7 +86,8 @@ export class CoqProject {
       this.coqProjectRoot = newSettings.coq.coqProjectRoot;
       this.console.log("Updated project root to " + this.getCoqProjectRoot());
     }
-    if(newSettings.coq.loadCoqProject ) {
+    if(newSettings.coq.loadCoqProject && !newSettings.coqtop.useDune) {
+      // ^^ CoqProject settings aren't needed when using dune
       this.watchCoqProject();
       await this.loadCoqProject();
     } else {
--- a/server/src/coqtop/CoqTop8.ts
+++ b/server/src/coqtop/CoqTop8.ts
@@ -222,8 +222,10 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
         // '/D /C', this.coqPath + '/coqtop.exe',
         '-main-channel', mainAddr,
         '-control-channel', controlAddr,
-        '-async-proofs', 'on'
-        ].concat(this.settings.args).concat(topfile);
+        '-async-proofs', 'on',
+        ...this.settings.args,
+        ...topfile
+      ];
     } else {
       var coqtopModule = this.coqtopBin;
       // var coqtopModule = 'cmd';
@@ -232,8 +234,9 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
         '-main-channel', mainAddr,
         '-control-channel', controlAddr,
         '-ideslave',
-        '-async-proofs', 'on'
-        ].concat(this.settings.args);
+        '-async-proofs', 'on',
+        ...this.settings.args
+        ];
     }
     this.console.log('exec: ' + coqtopModule + ' ' + args.join(' '));
     return spawn(coqtopModule, args, {detached: false, cwd: this.projectRoot});
--- a/server/src/coqtop/CoqTop8.ts
+++ b/server/src/coqtop/CoqTop8.ts
@@ -223,6 +223,7 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
   private spawnCoqTop(mainAddr : string, controlAddr: string) {
     var topfile : string[] = [];
     var scriptPath = this.scriptPath;
+    const relScriptPath = path.relative(this.projectRoot, scriptPath);
     if (semver.satisfies(this.coqtopVersion, ">= 8.10") && !this.settings.useDune && scriptPath != undefined) {
       // Dune computes this internally, so we skip it.
       topfile = ['-topfile', scriptPath];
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -56,10 +56,15 @@
...
+        "coqtop.dunePath": {
+          "type": "string",
+          "default": "dune",
+          "markdownDescription": "Full path to dune binary (`dune` included)."
+        },
         "coqtop.binPath": {
           "type": "string",
           "default": "",
-          "description": "Path to coqtop and coqidetop binaries (alternatively, hoqtop and hoqidetop)."
+          "description": "Path to folder with coqtop and coqidetop binaries (alternatively, hoqtop and hoqidetop)."
```
Blaisorblade, line 67: This is an unrelated message improvement.

package.json
```diff
@@ -85,6 +90,10 @@
           "default": "open-script",
           "markdownDescription": "When to start an instance of coqtop: when a Coq script is opened (`open-script`) or else when the user begins interaction (`interaction`; default)."
         },
+        "coqtop.useDune": {
+          "type": "boolean",
+          "markdownDescription": "Whether to use `dune coqtop`"
```
Blaisorblade, line 95: This description must be expanded, and the feature marked experimental.

Blaisorblade:
This isn't complete but I've been using it for a while (must still rebase over 0.3.7).

I'd like to ask what would be considered necessary. The lack of a decent progress dialog seems the most annoying problem...

Blaisorblade:
Bug report: this still passes flags from `_CoqProject` to Coq. That's one I can fix.

4ever2:
Doesn't this PR also fix #278?

Blaisorblade:
Yep, thanks @4ever2

package.json
Blaisorblade: Marked experimental. (Not sure what expansion I had in mind).

thery:
@Blaisorblade tell me when you think it is mergeable

huynhtrankhanh:
This PR is not ready to merge I guess. But one nitpick: why should we continue using `var`? We should use `const` and `let`, and prefer `const` whenever possible. `var` is function scoped, not block scoped like the other two keywords. This is very inconsequential for this PR though.

I'll probably review this PR more carefully later.

server/src/coqtop/CoqTop8.ts
```diff
@@ -227,16 +239,27 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
...
         '-async-proofs', 'on'
         ].concat(this.settings.args);
     }
-    this.console.log('exec: ' + coqtopModule + ' ' + args.join(' '));
-    return spawn(coqtopModule, args, {detached: false, cwd: this.projectRoot});
+    if (this.settings.useDune) {
+      var binary = this.settings.dunePath
+      if (scriptPath === undefined)
+        throw new CoqtopSpawnError("", "File was not saved to local file system");
+      var args = ["coq", "top", "--toplevel=" + coqtopModule,
```
huynhtrankhanh, line 254: Use `const`. Also use ES6 spread syntax instead of `concat`.

huynhtrankhanh (changes requested):

server/src/coqtop/CoqTop8.ts
```diff
@@ -210,15 +211,26 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
     return path.join(this.settings.binPath.trim(), this.settings.coqidetopExe);
   }
 
+  private get scriptPath() {
+    // TODO: rename scriptFile to scriptUri everywhere.
+    let uri = Uri.parse(this.scriptFile)
+    if (uri.scheme == "file")
+      return uri.fsPath;
+    else
+      return undefined
```
Blaisorblade, line 220: This is new since last revision and enables stepping through unsaved files (but only outside dune coq top).

server/src/coqtop/CoqTop8.ts
```diff
@@ -250,17 +250,16 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
         ...this.settings.args
         ];
```
Blaisorblade, line 251: I haven't ported this code to `const`, but it should be doable the same way. (But I prefer to avoid mixing refactorings with changes....).

server/src/coqtop/CoqTop8.ts
```diff
@@ -250,17 +250,16 @@ export class CoqTop extends IdeSlave8 implements coqtop.CoqTop {
...
+
+    if (this.settings.useDune && scriptPath === undefined) {
+      throw new CoqtopSpawnError("", "File was not saved to local file system");
     }
 
+    const [binary, args] = this.settings.useDune ?
+      [this.settings.dunePath, ["coq", "top", "--toplevel=" + coqtopModule,
+        scriptPath, "--", ...coqArgs]] :
+      [coqtopModule, coqArgs]
+
```
Blaisorblade, line 262: @huynhtrankhanh Since the code relied on `var`'s scoping, this seems the smallest possible change.

Had to lift error checking since TS seems to lack statement expressions, but I guess that's not too bad. I guess `function () { ... } ()` lets you encode them, but I wouldn't know if that's idiomatic.

Blaisorblade: To be clear, if anything seems silly it's because I'm not really a TypeScript dev and I was following existing idioms.

Blaisorblade:
I've done the main changes I planned, tried to address @huynhtrankhanh's comments, but I'll be retesting more later today.

server/src/coqtop/CoqTop8.ts
Blaisorblade: Done in last commit (untested yet).

Blaisorblade:
The latest changes seem to work, and I addressed @huynhtrankhanh 's comments, so I consider this ready for review and merging.

thery:
@Blaisorblade Perfect I will try to have a look at it today. Would be good if @huynhtrankhanh would give his opinion too

server/src/CoqProject.ts
```diff
@@ -86,9 +86,12 @@ export class CoqProject {
       this.coqProjectRoot = newSettings.coq.coqProjectRoot;
       this.console.log("Updated project root to " + this.getCoqProjectRoot());
     }
-    if(newSettings.coq.loadCoqProject ) {
+    if(newSettings.coq.loadCoqProject && !newSettings.coqtop.useDune) {
```
thery, line 89: does this means that if your flag is not set, everything works like as before?

thery (approved):

thery:
@Blaisorblade  I had a look, it seems ok.
As soon as @huynhtrankhanh is ok, we can merge

huynhtrankhanh:
This PR looks perfect. I really appreciate your work. However, there's one issue: the configuration options! There are two options here, which should be mutually exclusive: `coqtop.loadCoqProject` and `coqtop.useDune`. I believe having two mutually exclusive options that could be set independently causes problems when we further build open your work and makes it hard for users to predict the effect of their configuration. Should we have a single option that allows us to choose whether to load _CoqProject, use Dune or ignore both?

thery:
@huynhtrankhanh you would prefer a coqtop.configMode  with the options coqproject or dune?

server/src/CoqProject.ts
Blaisorblade: Yes, that was my design goal :-).

Blaisorblade:
Thanks for the review!

> Should we have a single option that allows us to choose whether to load _CoqProject, use Dune or ignore both?

Good idea; that works for me. I'm afraid I can get back to it in a couple days.

Dropping "ignore both" as @thery suggests also seems plausible but has a slightly higher chances of regressions — does any user _need_ to disable `loadCoqProject`? We should at least test projects with _no_ `_CoqProject` file.

thery:
@Blaisorblade @huynhtrankhanh I would like to make a new release beginning of march, do you think we can make this PR ready to merge by then?

Blaisorblade:
Sorry I'm late — I plan to address the comment next week, what I pushed just updates the extension for dune coq top changes (now dune wants relative paths, so we oblige).

thery:

Blaisorblade:
Sorry this hasn't progressed :-| ... Closing for now.

Issue #317, opened 2022-11-23
Issues with speed/output of goal display (1s in CoqIDE, 3min in VSCoq)


MSoegtropIMC:
Running over the following Coq snippet:
```
Require Import ZArith.

Open Scope Z.

Goal forall x, x>=15000.
intros x.
cbv.
```
takes about 1s in CoqIDE (until the result is displayed - CoqIDE does not block) and about 3 minutes in VsCoq (it blocks for this time).

Also the output is not the same - although I tried to ensure all display settings are the same (e.g. by restarting CoqIDE and VSCoq). For `10` instead of `15000` I get in CoqIDE:
```
1 goal
x : Z
______________________________________(1/1)
match x with
| 10 => Eq
| 0 | 7 | 9 | 5 | 3 | 6 | 8 | 4 | 2 | 1 | Z.neg _ => Lt
| _ => Gt
end = Lt -> False
```
and in VsCoq
```
x: Z
1/1
match x with
| Z.pos x' =>
    match x' with
    | (p~1)%positive =>
        match p with
        | (p0~1)%positive =>
            match p0 with
            | (p1~1)%positive | (p1~0)%positive => match p1 with
                                                   | (_~1)%positive | _ => Gt
                                                   end
            | 1%positive => Lt
            end
        | (p0~0)%positive =>
            match p0 with
            | (p1~1)%positive => match p1 with
                                 | (_~1)%positive | _ => Gt
                                 end
            | (p1~0)%positive => match p1 with
                                 | 1%positive => Lt
                                 | _ => Gt
                                 end
            | 1%positive => Lt
            end
        | 1%positive => Lt
        end
    | (p~0)%positive =>
        match p with
        | (p0~1)%positive =>
            match p0 with
            | (p1~1)%positive => match p1 with
                                 | (_~1)%positive | _ => Gt
                                 end
            | (p1~0)%positive => match p1 with
                                 | 1%positive => Eq
                                 | _ => Gt
                                 end
            | 1%positive => Lt
            end
        | (p0~0)%positive =>
            match p0 with
            | (p1~1)%positive => match p1 with
                                 | (_~1)%positive | _ => Gt
                                 end
            | (p1~0)%positive => match p1 with
                                 | 1%positive => Lt
                                 | _ => Gt
                                 end
            | 1%positive => Lt
            end
        | 1%positive => Lt
        end
    | 1%positive => Lt
    end
| _ => Lt
end = Lt -> False
```

[...]

PR #321, opened 2022-12-13
Add option to disable super-slow "Proof View Diff"

Blaisorblade:
- Some initial testing confirms @4ever2's diagnosis in https://github.com/coq-community/vscoq/issues/317#issuecomment-1347425838.
- Apologies if this is unfinished; this was a quick prototype, but it's already robust enough that one can enable and disable the option and see the effect immediately. And I know I don't really know TypeScript well.
- If I actually knew TypeScript, I'd maybe try using `SettingsState.proofViewDiff?` instead of adding a new boolean.

### Initial testcase

Adapted from https://github.com/coq-community/vscoq/issues/317#issue-1461682325 with size 500 (I didn't want to wait for 3 minutes).
```
Require Import ZArith.
Open Scope Z.

Goal forall x, x>=500.
intros x.
Time cbv.
```


server/src/stm/STM.ts
```diff
@@ -420,6 +420,9 @@ export class CoqStateMachine {
       endCommand();
     }
   }
+  private get proofViewDiff(): boolean {
+    return server.project.settings.coq.proofViewDiff
```
Blaisorblade, line 424: I initially thought this should be `this.project` instead of `server.project`, but the current code seems to mix both and hopefully they refer to the same state.

4ever2: It looks like they are the same, but it is probably safer to use `this.project` in case that doesn't continue to be the case.

Blaisorblade: That still worked!

4ever2: The rest looks good to me.

TheoWinterhalter:
I have a silly question: with your patch, can you still get proof diff but from Coq (which is faster if I understood correctly) or will this require yet another patch?

Blaisorblade:
@TheoWinterhalter it would need another patch, but I feel "VsCoq stops hanging for minutes" is still valuable, and a low-hanging fruit. The existing diffs still work AFAICT, and you can even tweak the setting without reloading the file.

thery:
Can we merge?

Blaisorblade:
Assuming the bump hasn't changed things, this should be pretty safe and keep the old behavior by default. (Maybe we should change the default).

Blaisorblade:
Thanks for merging!

PR #324, opened 2022-12-20
Fix formatting of ProofView error messages

### LINKED ISSUE #323: Newlines broken in ProofView error messages
It seems the latest changes have broken the error messages in the proofview.

This is the old behavior (at version 0.3.6) :
image

This is the new behavior (at commit 456b549a32a31784538ccb89adf7be43fb8efcbe) :
image

While the colors are a bit better the absence of line wrapping makes these error messages very illegible, especially so as environments get bigger.

4ever2:
Error messages in proof views are no longer formatted correctly as observed in #323.
This is broke in 4f9aacf071d19cbfa6b597646a05ae7e100311d2 due to missing attributes on the div containing the error message.

Fixes #323

### DIFF
--- a/html_views/src/goals/display-proof-state.ts
+++ b/html_views/src/goals/display-proof-state.ts
@@ -263,7 +263,7 @@ export const Infoview = () => {
       element.innerHTML = "";
       const formatted = h("div.message", [
         h("strong", "Error: "),
-        h("code", message),
+        h("code", {id: "error"}, message),
       ]);
       element.appendChild(formatted);
     };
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Thanks for the prompt bugfix!

I don't know yet the context, but maybe @huynhtrankhanh or @dlesbre can review or at least test? In doubt, we should consider merging this soon anyway.

huynhtrankhanh:
Sorry, this is me being careless :( Thanks!

PR #325, opened 2022-12-20
Fix proof view diff settings default values being ignored

4ever2:
Default values for the `coq.proofViewDiff.addedTextIsItalic` and `coq.proofViewDiff.removedTextIsStrikedthrough` are ignored after the addition of #321.
This was caused by overlapping setting keys.


Blaisorblade (approved):
Ouch, my bad. Thanks for the fix — merging!

PR #327, opened 2022-12-21
focus on the first goal when possible

thery:
This was in 0.3.6, don't understand why it was removed

### DIFF
--- a/html_views/src/goals/display-proof-state.ts
+++ b/html_views/src/goals/display-proof-state.ts
@@ -127,6 +127,7 @@ function createHypotheses(hypotheses: Hypothesis[]): HTMLElement {
 
 function createGoal(goal: Goal, index: number, goals: Goal[]): HTMLElement {
   const expressionElement = h("span.expr", createAnnotatedText(goal.goal));
+  if (index == 0) {expressionElement.setAttribute("id", "firstGoal")};
   const element = h("li", [
     h("span.goalId", `${index + 1}/${goals.length}`),
     expressionElement,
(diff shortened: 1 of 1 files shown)


huynhtrankhanh:
Yeah. This is because I rewrote the logic of the infoview and forgot to reimplement this feature. I wouldn't implement the feature this way though. We shouldn't add the firstGoal ID in the first goal. Instead, we should obtain a reference to the DOM element created in the `createFocusedGoals` call, and call `scrollIntoView` there. So a more idiomatic way to implement the feature would be this:
```typescript
const goalElements = createFocusedGoals(state.goals);
mainGoalsElement.appendChild(goalElements);
const firstGoalElement = goalElements[0];
firstGoalElement.scrollIntoView({ block: "center", inline: "nearest" });
```
Thanks for reminding me! I'll reimplement the feature.

thery:
As you've seen I am not an ts expert. I've tried your trick

thery:
My bad. This works
 ```typescript
 mainGoalsElement.appendChild(createHypotheses(state.goals[0].hypotheses));
      const goalElements = createFocusedGoals(state.goals);
      mainGoalsElement.appendChild(goalElements);
      const firstGoalElement = goalElements.firstElementChild;
      if (firstGoalElement) {
        firstGoalElement.scrollIntoView({ block: "center", inline: "nearest" });     mainGoalsElement.appendChild(createFocusedGoals(state.goals));
      }
```

thery:
> Another question, when I have several ProofView, I'd like that the one that receives a new goal to be visible. What is the proper way to do this?

@huynhtrankhanh any idea?

Blaisorblade:
> Another question, when I have several ProofView, I'd like that the one that receives a new goal to be visible. What is the proper way to do this?

See https://github.com/coq-community/vscoq/issues/82#issuecomment-1362983080.

PR #329, opened 2022-12-30
Improve hover provider

dlesbre:
I made a few small changes that improve the hover provider:

- **bugfix:** I noticed the hover provider wasn't rendering some notations. 
  This is because Coq omits parentheses for single word Notations. So:
  ```coq
  Notation foo := bar.
  About foo.
  ```
  Will print `Notation foo := bar` and not `Notation foo := (bar)`.

- **bugfix:** sometimes with implicit arguments the about query prints the type a second time 
  starting with `Expanded type for implicit arguments`. This broke `formatLocate` a bit as my split after type function would 
  split the output into three and not two.

- **Change:** I figure out how markdown formatting works so I could add some styling to the
  `Args:` and `Expands to:` sections of the hover.

- **Smallfix:** when in proof mode, instead of printing `Expands to: Variable foo`, `About foo` prints `Hypothesis of the goal context.`

### DIFF
--- a/client/src/HoverProvider.ts
+++ b/client/src/HoverProvider.ts
@@ -141,8 +141,8 @@ function formatAbout(response: string) {
     if (match !== null) {
       const notation = match[1].trim();
       let definition = response.slice(match[0].length).trim();
-      if (definition[0] == "(") {
-        const end = findClosingParenthese(definition, 0);
+      if (definition[0] === "(") {
+        const end = findClosingParenthese(definition, 1);
         if (end !== null) {
           definition = definition.slice(1, end);
         }
--- a/client/src/HoverProvider.ts
+++ b/client/src/HoverProvider.ts
@@ -122,13 +122,15 @@ function formatLocate(response: string) {
 
 // format the Expands to: xxx line returned by About queries
 function formatExpandsTo(line: string) {
-  const match_expands = line.match(/Expands to:\s+(\S+)\s+(.*)/);
+  const match_expands = line.match(/Expands to:\s+(\S+)(\s+.*)?/);
   if (match_expands !== null) {
     const md = new vscode.MarkdownString();
     md.appendMarkdown("**");
     md.appendText(match_expands[1]);
-    md.appendMarkdown("** ");
-    md.appendText(match_expands[2].trim());
+    md.appendMarkdown("**");
+    if (match_expands[2]) {
+      md.appendText(" " + match_expands[2].trim());
+    }
     return md;
   }
 
--- a/client/src/HoverProvider.ts
+++ b/client/src/HoverProvider.ts
@@ -131,6 +131,10 @@ function formatExpandsTo(line: string) {
     md.appendText(match_expands[2].trim());
     return md;
   }
+
+  if (line.trim() === "Hypothesis of the goal context.") {
+    return new vscode.MarkdownString("**Hypothesis**")
+  }
   return null;
 }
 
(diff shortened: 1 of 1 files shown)


thery:
@dlesbre Tell me when you think your PR is ready to be merged

client/src/HoverProvider.ts
```diff
@@ -128,42 +149,58 @@ function formatAbout(response: string) {
...
+  // |Notation z_to_addr := finz.of_z
+  // |Expands to: Notation cap_machine.addr_reg.z_to_addr
   if (response.match(/not a defined object\./gs) !== null) return;
 
   if (response.startsWith("Notation")) {
     const array = response.split(/\n(?!\s)/gms); // split on newline NOT followed by space
-    let hover = []
-    const match = array[0].match(/Notation\s+(.*?)\s+:=\s+\(/);
+    let hover: vscode.MarkdownString[] = []
+    const match = array[0].match(/Notation\s+(.*?)\s+:=\s*/);
```
Blaisorblade, line 161: I like your fix. I have only one question: I wonder if using this regexpr could make the code simpler:
```suggestion
    const match = array[0].match(/Notation\s+(.*?)\s+:=\s*(\()?/);
```
and then testing `match[2]`. Or maybe you could do the entire parsing with a regexp:
```suggestion
    const match = array[0].match(/Notation\s+(.*?)\s+:=\s*((\(?)[^)](\))?)/);
```

But this is just an idea, feel free to ignore it if it doesn't seem helpful.

dlesbre: I can use a regexp to figure out if the notation starts with a parentheses, but I can't do the entire parsing with a regexp since you can't parse well-parenthesized expression with regexps.

dlesbre: As it stands I think its more legible to have `definition[0] === "("` as the test than to have a more complex regexp

dlesbre:
As it stands the PR can be merged. 

However there is one last problem I'd like to fix but can't seem to figure out, which is how to properly wrap types.
As it stands the wrapping used is the one printed by `About`, which is based on the proofview's width. This render's ok if the proofview is smaller than vscode max hover width, but get very ugly if it isn't.

So instead of this:
```
∀ (Symbols : Type) (Symbols_eq_dec : EqDecision Symbols) 
  (Symbols_countable : Countable Symbols) (word_restrictions : 
                                           Word → 
                                           gset Addr → Prop) 
  (addr_restrictions addr_restrictions' : gset Addr → Prop),
  (∀ a : gset Addr, addr_restrictions a → addr_restrictions' a)
  → ∀ comp : component Symbols,
      well_formed_comp word_restrictions addr_restrictions comp
      → well_formed_comp word_restrictions addr_restrictions' comp
```

We get this :
```
∀ (Symbols : Type) (Symbols_eq_dec : EqDecision Symbols) 
  (Symbols_countable : Countable Symbols) 
(word_restrictions : 
                                           Word → 
                                           gset Addr → 
Prop) 
  (addr_restrictions addr_restrictions' : gset Addr → 
Prop),
  (∀ a : gset Addr, addr_restrictions a → 
addr_restrictions' a)
  → ∀ comp : component Symbols,
      well_formed_comp word_restrictions 
addr_restrictions comp
      → well_formed_comp word_restrictions 
addr_restrictions' comp
```

In theory this is fairly easy to fix: send a `Set Printing Width $VSCODE_HOVER_WIDTH` before the query, and then reset it after the query.

I have tried various ways sending `resize` to the server before performing the hover query but none had any visible effect... If you have any ideas why this is please let me know and I'll try again.

thery:
@dlesbre thanks!
Ps : Open an issue about your width problem so we don't forget it.

PR #335, opened 2023-02-02
[sel] priority among tasks

gares:
This PR packs some orthogonal features (hem...) but I don't think it is worth making the history cleaner at this stage.

- sel now takes a `todo` which internally keeps system/queue/task events separate (for my mental sanity)
- tasks have now a priority, and are always returned accordingly (pretty naive, but that is it)
- events can be named, for debugging purposes
- events can be made recurrent (a copy always stays in the queue, we have two occurrences in the lang server, the lsp message event, and the some feedback from coq one)

The library is still a little low level, e.g. if you have two `ExecToLoc` events (tasks) it returns both, it is up to the caller to
give precedence to the first one, and put the other back (with the same priority, likely).
This is a deliberate choice, and lets one for example overrule both `ExecToLoc` if something happens, or purge/arrange one of the two if the other one includes/overlaps with it. I think these scheduling decision are not to be done by SEL.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
       run: opam install ./language-server/vscoq-language-server.opam --deps-only --with-doc --with-test
 
     - name: Install vscoq-language-server
-      run: opam install ./language-server/vscoq-language-server.opam
+      run: opam install ./language-server/vscoq-language-server.opam --with-test
 
     - run: |
         eval $(opam env)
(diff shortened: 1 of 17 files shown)


gares:
CI is broken on nix, no idea

gares:
I'm not happy with the names of the APIs, but it should do what we discussed @maximedenes .
In particular, as a test checks, you can now kill a task by starvation but making a recurring, higher priority.

PR #338, opened 2023-02-03
Add missing snippet placeholders

### LINKED ISSUE #112: "Arguments" snippet of autocompletion more annoying than useful
Perhaps I'm missing something but the autocompletion for `Arguments` yields:
```coq
Arguments {{qualid}} {{possibly_bracketed_idents …}}.
```
It does not even select the bracketed strings that you have to modify. As far as I'm concerned, merely completing `Arg...` to `Arguments` would be more practical.

The `Definition` snippet behaves in a different way so I suspect this one is a bug.

4ever2:
Some snippets like `Set`, `Print`, and `Arguments` were not using placeholders for their arguments, making the snippets annoying to use.

Fixes #112


thery:
Thanks

PR #364, opened 2023-02-08
[opam] fix packaging

gares:
TODO:
- [ ] fix nix since submodule is gone


gares:
@maximedenes let's wait the e2e test before merging.
I hacked something around with-test, since with-dev-setup is broken/not-documented/who-knows, so that we have one opam file which is ready for distribution, and we can also use it with our submodule setup.

PR #367, opened 2023-02-08
fix: Nix flake target error in VSCoq2

Duckapple:
To build the nix flake, @HjalteSorgenfrei, @Jakobis and I needed to add that it was the default target for linux specifically.
We honestly don't know enough about Nix to know if `packages.default` also should be provided, so if something here is wrong, then feel free to correct us.

That would probably also solve [the failing action](https://github.com/coq-community/vscoq/actions/runs/4125391653/jobs/7125890414).

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,4 +4,4 @@
     sha256 = "0jm6nzb83wa6ai17ly9fzpqc40wg1viib8klq8lby54agpl213w5"; }
 ) {
   src =  ./.;
-}).defaultNix.packages.default
+}).defaultNix.packages.x86_64-linux.default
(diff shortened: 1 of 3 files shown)


maximedenes:
Oops, sorry for breaking this. I'll use PRs systematically now that we have a working CI. You can remove `packages.default`, it was actually a mistake.

Duckapple:
As it turns out, when I entirely replaced `packages.default`, then nix-shell command failed, so it would probably need fixing in there too.

maximedenes:
You should probably update `shell.nix` in the same way you did `default.nix`.

maximedenes (approved):
Thanks a lot and sorry for breaking this!

PR #370, opened 2023-02-09
[VSCoq2] Show a message when not in proof or no more subgoals available

Duckapple:
An empty panel felt like something has gone wrong, so I added a message to let you know that there are no more subgoals, or that you are not in a proof (much like the view in VSCoq 1 used to show.)
I introduced a property `isInProof: boolean` to the goal state response and made the LSP return `{ isInProof: false }` when the document manager can't find a proof at the current location.

I'd love to know if I have misstepped in my code style.

### Screenshots

Shows "no more subgoals"

Shows "not in proof"


language-server/vscoqtop/lspManager.ml
```diff
@@ -257,7 +257,15 @@ let coqtopUpdateProofView ~id params =
   let loc = params |> member "position" |> parse_loc in
   let st = Hashtbl.find states uri in
   match Dm.DocumentManager.get_proof st loc with
-  | None -> ()
+  (* Send an "empty" response to notify about not being in a proof *)
+  (* TODO: Selectively updates based on prior state (should not alert when already not in proof) *)
```
maximedenes, line 261: Actually, I disagree with the `TODO` (i.e. I think your current approach is fine) :)
The protocol should be stateless, it is the job of the UI to be smart about how it notifies IMHO.

.gitignore
```diff
@@ -1,4 +1,5 @@
 client/node_modules
 client/dist/
+client/out/
```
maximedenes, line 3: Out of curiosity, what are we producing in `out`?

maximedenes (commented):
Thanks a lot for this nice improvement!

.gitignore
rtetley: From the package.json it's for the compile tests. This is actually from the vscode defaults, at some point we should look into it more ?

client/src/protocol/types.ts
```diff
@@ -18,6 +18,7 @@ export interface UpdateProofViewRequest {
 }
 
 export interface UpdateProofViewResponse {
+    isInProof: boolean;
```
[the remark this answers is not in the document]
rtetley: Actually  I was looking at the code and thinking that their is actually two separate empty states here. I don't know if you follow me but it feels to me that: 

1) If we're not in proof mode, we shouldn't try to render organisms at all (i.e GoalTabs or GoalCollapsibles), 
so in the GoalPage we should have the isInProofMode check, something like: 
`const pageBody = isInProofMode ? <GoalTabs /> : 'Not in proof mode'`
2) If we're in proof mode, it's the organisms job to check if the goal has been completed or not, it has it's own emtpy state, so in the relevant organism: 
`const goalSection = goals.length ? goals.map(blabla) : emptyStateString 

Does that make sense ?

rtetley:
Thanks for the feature, the coding style looks fine on the front end side of things :-)

language-server/vscoqtop/lspManager.ml
Duckapple: That is a fair point, I'll add it to my review fixes.

client/src/protocol/types.ts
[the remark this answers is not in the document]
Duckapple: I have made the decision flow a little more clear in my review fixes.

maximedenes:
I'd like to make the types slightly more precise, then it should be ready for merge. I'll give it a shot asap.

rtetley:
I would like to do a bit of refactoring in the front end code. Will do ASAP as well.

Issue #380, opened 2023-02-14
Declaration Provider for VsCoq2


Hjaltesorgenfrei:
A declaration provider would make it possible to navigate to the implementation of declarations in Coq.
VsCode has a registration for them <https://code.visualstudio.com/api/references/vscode-api#DeclarationProvider>

We have a draft implementation on <https://github.com/Duckapple/vscoq/tree/vscoq2-itu-develop>

rtetley:
Closing in favour of #214

PR #381, opened 2023-02-14
[VsCoq2] Add naive autocomplete and navigate to declaration

### LINKED ISSUE #377: Simple auto-suggestion for VSCoq 2
As a user, I would like to have suggestions for lemmas and hypotheses in VSCoq using the VSCode inline completion interface.

We already have this simple suggestion feature done for all lemmas in scope, as well as hypotheses for the current goal.
Filtering is provided via VSCode default filtering, which means it will filter on the start of the name.

PR is incoming.

Hjaltesorgenfrei:
fixes #377 and #380.
Also adds a setting for Coqlib so you do not have to use a command line argument for it.

A more thorough review of our ocaml files by @maximedenes would be nice, as we are bit a unsure about how we should structure the files and where what part of the implementation should be.  

If you want we can also separate Go To Declaration to it's own PR.


Hjaltesorgenfrei:
This is currently wrong, we are using the wrong method for hover and autocomplete and should be using the ones from the spec.
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

Hjaltesorgenfrei:
> This is currently wrong, we are using the wrong method for hover and autocomplete and should be using the ones from the spec. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

This is now fixed and we conform to the spec for completion and go to declaration

maximedenes:
Ouch, changing the base branch closed the PR without notice :(

PR #386, opened 2023-02-15
Fix JSON encoding of `Range.t`

maximedenes:


gares:
was this causing a slowdown?

gares:
I mean, also a slowdown?

maximedenes:
No, it just broke lsp decoding of elements containing ranges.

PR #388, opened 2023-02-15
Add naive autocomplete

### LINKED ISSUE #377: Simple auto-suggestion for VSCoq 2

maximedenes:
Reopens #381 by @Hjaltesorgenfrei after main branch changed.
Original description:

fixes #377 and #380.
Also adds a setting for Coqlib so you do not have to use a command line argument for it.

A more thorough review of our ocaml files by @maximedenes would be nice, as we are bit a unsure about how we should structure the files and where what part of the implementation should be.

If you want we can also separate Go To Declaration to it's own PR.


maximedenes:
@Hjaltesorgenfrei sorry for the mess due to the change of base branch. I fixed the build (some LSP encoding API changed in the server). I also had a look at the content of the PR:
- I'm not sure the `coqlib` setting is a good idea: the hope was that the need for this parameter was temporary. There are more systematic ways to know where Coq libraries are installed when they are. The tricky part is that for now we are building Coq locally as part of the dev build of VsCoq2. But passing an extra flag is ok for developers.
- In the jump to declaration part, you seem to assume that `.v` are located in the same folder as their corresponding `.vo`, right? Unfortunately, this is not going to be true for many projects. In fact, "hygienic" builds tend to produce artifacts in separate folders, as it is a best practice.
- The autocompletion part looks like a good first step to me. How about extracting that part in a separate PR that we can merge right away?

Hjaltesorgenfrei:
I think that is good idea, i will extract the declaration navigation out into a separate branch.

Hjaltesorgenfrei:
@maximedenes Are we waiting on #392 and doing the capability in the same way here?

maximedenes:
Sorry, I saw your comment after merging. #392 is in `main` now, it cleans up a bit the capabilities encoding so you can follow the same pattern.

Hjaltesorgenfrei:
I have added the capabilities type now, but I did not add anyway to convert the Json back. My reasoning being that the server should not be receiving server capabilities and therefore never deserializing.

language-server/lsp/lspData.ml
```diff
@@ -91,9 +104,36 @@ module ServerCapabilities = struct
...
   | _ -> Yojson.json_error "invalid value"
 
+  type completionOptions = {
+    triggerCharacters : string list option;
+    allCommitCharacters : string list option;
+    resolveProvider : bool option;
+    completionItemLabelDetailsSupport : bool option;
+  } [@@deriving yojson]
+
+  let yojson_of_completionOptions options =
```
maximedenes, line 114: Sorry, I should have explained better: you don't need to write this function anymore. The `[@@deriving yojson]` annotation you put on the type derives it automatically.

Hjaltesorgenfrei: The spec does not accept a null value: <https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionOptions> 
And it is my understanding yojson would decode a option to null instead of leaving it out. 
So should we remove the deriving, the values we don't use (leaving only resolveProvider) or the option and then use the correct values for each.

maximedenes: You are right. We should add `[@yojson.option]` to optional fields, to strictly obey the spec.

maximedenes: Do you mind adding it where appropriate? It will make `None`-valued fields be omitted rather than set to `null`.

Hjaltesorgenfrei: Yep, i will add it.

language-server/vscoqtop/lspManager.ml
```diff
@@ -211,6 +216,23 @@ let coqtopStepForward ~id params : (string * Dm.DocumentManager.events) =
   Hashtbl.replace states uri st;
   update_view uri st;
   (uri,events)
+  
+  let make_label (label, typ, path) =
```
maximedenes, line 220: How about defining this type (in `LspData`) and letting `yojson` do the encoding? The "Path" concatenation could be done outside the encoding part, for better separation of concerns.

Hjaltesorgenfrei: Yeah, thats a better idea. I will change it later today

Hjaltesorgenfrei:
I fixed the completion item serialization and changed it to CompletionList, which should allow us to send an incomplete list and refine it when the user types more. This should allow us to reduce the amount sent to the user.

maximedenes:
Thanks a lot! I can take care of the conflicts and merging.

PR #395, opened 2023-02-16
Handle the check query in the backend

### LINKED ISSUE #350: Support for `Check` & `About` in query panel

rtetley:
… the front.

Closes #350.

### DIFF
--- a/language-server/dm/documentManager.ml
+++ b/language-server/dm/documentManager.ml
@@ -249,12 +249,12 @@ let get_proof st pos =
   | Some sentence ->
     ExecutionManager.get_proofview st.execution_state sentence.id
 
-  let get_context st pos =
-    let loc = Document.position_to_loc st.document pos in
-    match Document.find_sentence_before st.document loc with
-    | None -> None
-    | Some sentence ->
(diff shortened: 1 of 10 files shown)


rtetley:
The types for the three currently supported queries are very similar. Maybe we should refactor them into a single type ? Or should we stay explicit ? Will they get different in the future ?

maximedenes (approved):

language-server/dm/documentManager.ml
```diff
@@ -289,6 +289,43 @@ let search st ~id pos pattern =
...
+    Prettyp.print_judgment env sigma j ++
+    Printer.pr_ne_evar_set (fnl () ++ str "where" ++ fnl ()) (mt ()) sigma l
+  in
+  pp ++ Printer.pr_universe_ctx_set sigma uctx
+
+  let check st pos ~goal ~pattern =
+    match get_context st pos with 
+    | None -> Error ("No context found") (*TODO execute *)
+    | Some (env, sigma) ->
+      let rc = parse_entry st pos Pcoq.Constr.lconstr pattern in (* TODO parse contr_expr instead of smart_global *)
```
maximedenes, line 322: TODO is done here

PR #398, opened 2023-02-17
[VSCoq2] fix: Hover now returns Null when not hovering a word.

Duckapple:
[The LSP documentation implies that the language server should return `contents: null` if no hover text should be showed.](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover)

I found that the client would pop-up a hover with the contents "Loading..." whenever the mouse rested on whitespace and the like.

With this fix, we now respond that we can't perform a hover at that point, making no hover popup appear.

@rtetley, should I assign you as reviewer/assignee?

### DIFF
--- a/language-server/dm/documentManager.ml
+++ b/language-server/dm/documentManager.ml
@@ -291,10 +291,7 @@ let search st ~id pos pattern =
 
 let hover st pos = 
   let opattern = Document.word_at_position st.document pos in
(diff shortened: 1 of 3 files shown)


rtetley:
Thanks for the correction ! Sure I'll review the code !

rtetley (approved):

PR #399, opened 2023-02-17
Show a message when not in proof or no more subgoals available

### LINKED ISSUE #369: Show message when all goals are solved

maximedenes:
Reopens #370 after change of base branch.

Fixes #369.


rtetley:
@Duckapple I have a few modifs that I'd like to push. Could you click on "allow edits and access to secrets by maintainers" ? Then I'll push and we can merge !

Duckapple:
I went to the old PR and ticked the box, hope that helps

rtetley:
Nope still gets rejected :-/ if you can't do it on this PR maybe @maximedenes can ? Although I doubt it. In case we might have to open yet another PR -__-

maximedenes:
> I went to the old PR and ticked the box, hope that helps

Doesn't GitHub show the box for this PR?

rtetley:
While this is solved I pushed my modifs to another branch with the same name here

client/search-ui/src/utilities/hooks.ts
```diff
@@ -0,0 +1,24 @@
...
+    useEffect(() => {
+      // cb.current is `undefined` on initial render,
+      // so we only invoke callback on state *updates*
+      if (cbRef.current) {
+        cbRef.current(state);
+        cbRef.current = undefined; // reset callback after execution
+      }
+    }, [state]);
+  
+    return [state, setStateCallback];
```
Duckapple, line 23: What is this for?
If you wanted a version of `useState` where you want the previous state, then the following use case should be enough:
```ts
const [state, setState] = useState(false);
setState((prev) => !prev);
```

Duckapple:
@rtetley I merged your branch into mine. I looked through the code, and my only comment is the above about your newly introduced `useStateCallback`

client/search-ui/src/utilities/hooks.ts
rtetley: @Duckapple Sorry about that one ! It wasn't meant to be pushed ! 

So before React introduced hooks, you could put a call back when setting a state that would be called AFTER the state was modified. This is very useful when you want to be able to setState after another setState, or do anything that is dependent on the new state. 

Nowadays with hooks, you can have a similar thing by using the useEffect hook. However, this useEffect will always be called when the given arguments are modified, which includes even the initial render. This is very unpractical and can cause some havoc. Also it was really handy to be able to call a different callback contextually. 

This hook behaves similarly to the old class setState call back. For reference it comes from here: 
https://stackoverflow.com/questions/54954091/how-to-use-callback-with-usestate-hook-in-react

In our case I was thinking of using it for some functionalities such as saving the state in user workspace (vscode.saveState). 

However I will debug the panel which is currently wonky in a different branch and different PR. We can remove it from this PR.

Duckapple: Ah, I see... I didn't know that behaviour existed back then, that's kinda neat!

PR #409, opened 2023-02-19
[VsCoq1] Fix getSentenceIndexBeforeOrAt()

4ever2:
The function `getSentenceIndexBeforeOrAt()` returns `-1` when given a position that is inside the last sentence in the document rather than returning the index of the last sentence.
This results in completion suggestions failing at random.

### DIFF
--- a/client/src/Snippets.ts
+++ b/client/src/Snippets.ts
@@ -285,7 +285,8 @@ export function setupSnippets(subscriptions: vscode.Disposable[]) {
   subscriptions.push(vscode.languages.registerCompletionItemProvider('coq', {
     provideCompletionItems: async (doc, pos, token) => {
       try {
-        const prefix = await CoqLanguageServer.getInstance().getPrefixText(doc.uri.toString(),pos,token);
+        const prefix = await (await CoqLanguageServer.getInstance().getPrefixText(doc.uri.toString(),pos,token)).trimStart();        
+
         if(prefix === "")
           return [];
         const trigger = getTriggerSnippet(prefix);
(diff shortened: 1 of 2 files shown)


thery:
Could not reproduce the bug.  I've tried to do auto-completion while being on the last line. Could you be more precise?
Thanks.

4ever2:
Yes, I should have been more clear about what this PR fixes. It fixes three different bugs related to the completion suggestions provided by VsCoq. There is a distinction between auto-completion in general and the suggestions provided by VsCoq.
In general, there are three kinds of auto-completion suggestions that you will see when editing Coq files in VsCoq:
1) Completions based on keywords present in the file. These are provided directly by VS Code and not by the extension.
2) Snippets provided by VsCoq in https://github.com/coq-community/vscoq/blob/vscoq1/client/snippets/coq.json
3) More advanced snippets (like the `Set`, `Hint`, and `Print` snippets) provided by VsCoq in https://github.com/coq-community/vscoq/blob/vscoq1/client/src/Snippets.ts

The three bugs are only related to the last type of auto-completion suggestion.
To show these bugs consider the following file
```

Definition a := 0.

Definition b := 0.

Definition c := 0.

```
1) If you insert a space on the first empty line, you should see a list of snippet suggestions. However, getting auto-completion suggestions when entering a space doesn't make sense and is very annoying when writing proofs.
2) If you start writing `Hint` on the first line you should see a `Hint...` auto-completion. However, if you enter the same on the last line you will not get it.
3) If you start typing `Hint` the lines between definitions `a` and `c` you should see that the `Hint...` suggestion sometimes works and other times doesn't.

thery:
@4ever2  Thanks!

PR #413, opened 2023-02-20
Continuous / manual checking mode

### LINKED ISSUE #344: Continuous & manual checking modes
Provide settings to switch between modes:
- Full continuous checking
- Manual checking (requires "Interpret to point" and other navigation commands + view the checked zone)
- Manual checking with electric terminator

rtetley:
Closes  #344
- [x] create option in package.json
- [x] handle proof.mode option parsing in backend 
- [x] handle continous mode / manual mode in backend : 
    - [x] handle interpret to point 
    - [x] handle interpret to end
    - [x] handle step forward
    - [x] handle step back
- [x] Re-wire the updateProofView request

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -113,6 +113,16 @@
         "title": "Proof checking",
         "type": "object",
         "properties": {
+          "vscoq.proof.mode": {
+            "scope": "window",
+            "type": "string",
+            "enum": [
+              "Manual",
+              "Continuous"
+            ],
+            "default": "Continuous",
+            "description": "Configures the proof checking mode for Coq."
+          },
           "vscoq.proof.delegation": {
             "scope": "window",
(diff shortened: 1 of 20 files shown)


rtetley:
Since #427 we get the conf in the backend. I think we should merge this and open new issues for stepForward and stepBackward. Up to you @maximedenes !

maximedenes:
@rtetley Can you rebase so that we see only the relevant commits?

rtetley:
Just to close the conversation, the commits regarding the configuration initialization are in this branch because I needed the server to get the correct mode option (continuous or manual) and only realized here that the workspace/configuration request does not do the trick. 

Also regarding the stepForward and  stepBack we ended up implementing it with @maximedenes . 

We should merge at this stage !

Issue #379, opened 2023-02-14
Inline suggestions for VSCoq 2


Duckapple:
As part of our Master's thesis, @Hjaltesorgenfrei, @Jakobis and I want to improve Coq with inline suggestions/completion items.

The idea is to bring Coq "up to standard" with other mainstream/general-purpose programming languages in terms of providing suggestions for lemmas, tactics, goal hypotheses, etc., and also filter and sort the suggestions such that they are relevant to the current context. 

This issue will be long-standing as we work on different aspects of providing these suggestions, and we intend to make the changes in shorter-lived PRs and issues. 
We will link this issue with other issues and PRs we submit, so we can keep a thread over the larger narrative.

Currently, as described in #377, we have some suggestions already implemented. 

Below I have attached an image of our current prototype, in case it is not obvious what feature we are implementing:
![image](https://user-images.githubusercontent.com/39386945/218723148-1091c308-10cc-4f27-aa7a-248f857faf28.png)

Duckapple:
@maximedenes If I understood correctly, did you want this issue on the project board?
If yes, you will probably want to add it for me, I don't have the permissions :)

PR #416, opened 2023-02-22
Rank possible completions

Hjaltesorgenfrei:
Work to improve #379.
Ranks completion items based upon atomic type overlap.
It gets the atomic types of goal and lemmas, then checks for the largest overlap between the lemma and goal. 
This is first done by seeing which has most types in common with the goal, if they have an equal amount, then the smallest set of types is used. By doing it this way it should show lemmas that has the most possible types that are relevant to the current goal, while not showing gigantic lemmas.

For a very simple lemma, it suggests the correct lemma as the first one.
image

A draw back with this method is that it does not show very general lemmas, as they will have a very small overlap.

@gares @maximedenes do you think this method of ranking the items is sound?

### DIFF
--- a/language-server/dm/completionSuggester.ml
+++ b/language-server/dm/completionSuggester.ml
@@ -1,5 +1,33 @@
 module CompactedDecl = Context.Compacted.Declaration
 open Printer
+open EConstr
+open Names
+
+type unifier = 
+  | SortUniType of ESorts.t
+  | AtomicUniType of types * unifier array
+
+module UnifierCompare = struct
+  let evd = ref (Evd.from_env (Global.env ()))
+  type t = unifier
+  let rec compare l r = match l, r with
+    | SortUniType s1, SortUniType s2 -> if ESorts.equal !evd s1 s2 then 0 else 1 (* 1 or -1 does not really matter *)
+    | SortUniType _, _ -> 0
+    | _, SortUniType _ -> 0
+    | AtomicUniType (t1, ua1), AtomicUniType (t2, ua2) -> 
+      let c = EConstr.compare_constr !evd (EConstr.eq_constr !evd) t1 t2 in
+      if not c then 1 (* 1 or -1 does not really matter *)
+      else if Array.length ua1 = Array.length ua2 && Array.for_all2 (fun l r -> compare l r = 0) ua1 ua2 then 0 else 1
+end
+
+module Unifiers = Set.Make(UnifierCompare)
+
+module TypeCompare = struct
+  type t = types
+  let compare = compare
+end
+
+module Atomics = Set.Make(TypeCompare)
 
 let mk_hyp sigma d (env,l) =
   let d' = CompactedDecl.to_named_context d in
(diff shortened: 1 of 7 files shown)


gares:
Looks reasonable.

I think it could be simplified by using "sets of GlobRef.t" rather than sets of "types" which are arbitrarily complex, and maybe too precise (I guess `list A` and `list nat` don't overlap at all, while `{list,Rel -1}` and `{list,nat}` would overlap).

Here some code you can look at
https://github.com/LPCIC/coq-elpi/blob/efcf63c433834592beb221c09a5e76f9a847128a/src/coq_elpi_builtins.ml#L1105-L1132
(it filters stuff in a given module, maybe not immediately relevant to you, but uses the simplest API for doing it).

gares:
Frankly, I'm quite bluffed by you guys, the APIs of Coq are super messy, and still you write code which makes a lot more sense of these APIs ;-)

language-server/dm/completionSuggester.ml
```diff
@@ -26,11 +105,27 @@ let get_hyps st loc =
...
   let open Yojson.Basic.Util in
   let hypotheses = get_hyps st loc in
   let lemmasOption = DocumentManager.get_lemmas st loc in
-  let lemmas = lemmasOption |> Option.map (List.map CompletionItems.pp_completion_item) in
-  [hypotheses; lemmas] 
+  let goal, sigma, env = get_goal_type st loc in
+  let lemmas = lemmasOption |> Option.map 
+    (fun l -> 
+      rank_choices goal sigma env l |> 
+      take 10000 |>
```
gares, line 126: Did you try if vscode can really handle that many results? (even faking the list).
Cause it looks like *a lot* to me.

Hjaltesorgenfrei: It might be a bit excessive, I just set it to 10k so that our `sortText` works right now, as we are using 5 digit numbers to sort the results with. 
We will figure out a more reasonable limit at some point.

Hjaltesorgenfrei:
Thanks for the feedback and link :)

I am not sure how a `GlobRef.t` changes what is compared, how would comparing them be different to using types?
In the case of lists I would assume it is still just a reference to the list type?
 
But i do see your point about `{list, Rel 1}` and `{list, nat}`.
```coq
(* Atomics: [@eq,list,@rev,_UNBOUND_REL_2,_UNBOUND_REL_1] *)
Theorem rev_involutive : forall X : Type, forall l : list X,
  rev (rev l) = l.


Theorem rev_exercise1 : forall (l l' : list nat),
  l = rev l' ->
  l' = rev l.
Proof.
  intros **.
  symmetry.
  rewrite H. (* Atomics: [@eq,nat,list,@rev,l'] *)
  apply rev_involutive.
Qed.
```
At the point where `rev_involutive` is applied, the set of atomic types contains nat and actually prefers suggesting nat related lemmas as compared to suggesting `rev_involutive`.
So unpacking the types completely is a bad idea and we will change it to support polymorphism.

PR #419, opened 2023-02-23
Cleaning the panels

### LINKED ISSUE #402: Debugging the query panel

rtetley:
The panels have certain wonky behaviors that need to be cleaned. This also calls for some refactoring. Closes #402  
- [x] Refactoring search-ui panel 
- [x] Debugging weird tab selection behaviour 
- [x] Debugging state cycle in search ui 
- [ ] Solution for race condition when calling query command from the text editor


rtetley:
I am still running in two problems: 
- [ ] delete handler bug -> currentTab does not want to update 
- [ ] immediateQueryNotification -> still not working from a closed panel initial state

maximedenes (approved):

PR #422, opened 2023-02-27
Keybindings for queries + check command

rtetley:
Just added the relevant keybindings. Awaiting merge of #419  and #413  for clean up and merge.


rtetley:
Indeed ;-) Fixed !

PR #429, opened 2023-02-28
log initialization to a file

gares:


language-server/vscoqtop/vscoqtop.ml
```diff
@@ -46,20 +46,34 @@ let vscoqtop_specific_usage = {
   extra_options = "";
 }
 
+let init_log =
+  (* we put the time to make the file names easy to sort *)
+  try Some (open_out @@ Filename.temp_file (Printf.sprintf "vscoq_init_log.%f." (Unix.gettimeofday())) ".txt")
```
maximedenes, line 51: I don't think it is useful to put the time in the name, filesystems already expose it, don't they?

gares: I never recall how to sort by creation time with ls, while I recall I can do ls|sort

maximedenes: `ls -ltr` ;)

gares, line 51:
```suggestion
  try Some (open_out @@ Filename.temp_file "vscoq_init_log." ".txt")
```

gares:
@maximedenes ping

PR #438, opened 2023-03-09
[VsCoq2] Move try statement as parsing entry might fail

Hjaltesorgenfrei:
Hovering some words, such as "forall" made the server crash. If we move the try statement it should crash less


gares (approved):

gares:
Thanks!

PR #439, opened 2023-03-09
Coq bin detection

rtetley:


client/src/utilities/toolchain.ts
```diff
@@ -0,0 +1,88 @@
...
+            return this.searchForVscoqtopInPath();
+        }
+
+    }
+
+    private searchForVscoqtopInPath () : string {
+        const pathVars = this.splitEnvPath(this.getEnvPath());
+        VsCoqToolchainManager._channel.appendLine('Searching Path var: ');
+        for(let i in pathVars) {
+            if(path.basename(pathVars[i]) === 'vscoqtop') {
```
gares, line 66: Maybe its just my js which sucks, but I was expecting to find something like
```
for(p in paths) { if (file_exists(p + delimiter + "vscoqtop")) { alleluia } }
```
but here TTBOMU you look for a path entry whose last component is vscoqtop, while PATH contains the directories in which the binary vscoqtop may be (or not).

rtetley: You are totally right ! I was clearly a bit tired yesterday !

client/src/utilities/toolchain.ts
```diff
@@ -0,0 +1,90 @@
...
+        for(let i in pathVars) {
+            if(await isFileInFolder('vscoqtop', pathVars[i])) {
+                return pathVars[i];
+            }
+        }
+        return "";
+    }
+
+    private checkIfCoqFound(vscoqtopPath: string) {
+        const cmd = "./vscoqtop -where";
```
gares, line 77: you should also pass the options coming from user settings I believe, eg -coqlib

rtetley: Done. However the command still fails !

rtetley:
@maximedenes @gares is this working as intended, as I recall we were waiting for maxime to rebase the submodule but it should be all right now ? if so let's merge !

client/src/utilities/toolchain.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    private checkIfCoqFound(vscoqtopPath: string) {
+        const options = workspace.getConfiguration('vscoq').get('args') as string[];
+
+        const cmd = ["./vscoqtop -where"].concat(options).join(' ');
+        VsCoqToolchainManager._channel.appendLine("Launching command: " + cmd + " in folder: " + vscoqtopPath);
+        exec(cmd, {cwd: vscoqtopPath}, (error, stdout, stderr) => {
+            if(error) {
+                VsCoqToolchainManager._channel.appendLine(error);
+                window.showErrorMessage(stderr);
```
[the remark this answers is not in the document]
rtetley: Done. I was wonder, if for whatever reason the user passes -coqlib and the dir does not work, the server will crash without real warning (-where does not detect that, it ignores the -coqlib arg). Is there a way to check that and notify the user that the path he gave is incorrect ?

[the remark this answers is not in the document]
gares: Starting from 8.18 coq* -where will honor -coqlib

PR #444, opened 2023-03-13
[VsCoq 1] Fix #443

### LINKED ISSUE #443: [VsCoq 1] End section color bug
This is new in vscoq 0.3.8, the `End Foo` command is now always colored as an error:

image

I think this is due to some code to highlight `end` in red when no `match` is found, as adding an incomplete `match` statement fixes this:

image

dlesbre:
Remove matching pairs of words from languages bracket configuration. VS Code doesn't seem to like words there (matching is case insensitive, as far as I can tell the `Proof.` matches never matched either...).

### DIFF
--- a/client/coq.configuration.json
+++ b/client/coq.configuration.json
@@ -21,30 +21,6 @@
     [
       "(",
       ")"
-    ],
-    [
-      "Proof.",
-      "Qed."
-    ],
-    [
-      "Proof.",
-      "Admitted."
-    ],
-    [
-      "Proof.",
-      "Abort."
-    ],
-    [
-      "match",
-      "end"
-    ],
-    [
-      "repeat match",
-      "end"
-    ],
-    [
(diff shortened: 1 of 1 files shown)


4ever2:
Adding module and section to bracket configuration also fixes the problem.
```
[
  "section",
  "end"
],
[
  "module",
  "end"
]
```

It seems that any bracket pattern including uppercase or `.` is broken.

dlesbre:
True but I'm not sure we want to keep them for multiple reasons:

- The case insensitivity means you can get wrong highlights
- You can get wrong highlight because section/module names also need to match
- Getting it to work on proofs is very hard because the `Proof` isn't required and often skipped in things like obligations...

I agree the full feature (red color for badly parenthesized proofs, section, modules, etc...), I'm just not certain this is the right way to do it. Ignoring the problem with proof, this should be manageable with a basic parser though.

PR #446, opened 2023-03-14
[vscoq2] cleanup options data type and logics

gares:


maximedenes (dismissed):
After discussion: maybe better to put a ref in execution manager and then switch to an (explicit) global state in exec manager, instead of per-doc.

gares:
done

gares:
Damn, I need a green light to merge. @maximedenes would you unblock this swiftly?

PR #448, opened 2023-03-14
[vscoq2] move doc_id inside execManager

gares:
feedback handling is still not correct

### DIFF
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -44,15 +44,23 @@ type options = {
 }
 let default_options = { delegation_mode = CheckProofsInMaster }
 
+let doc_id = ref (-1)
+let fresh_doc_id () = incr doc_id; !doc_id
+
 type state = {
   initial : Vernacstate.t;
   of_sentence : (sentence_state * feedback_message list) SM.t;
+  doc_id : int; (* unique number used to interface with Coq's Feedback *)
+
 }
 
-let init vernac_state = {
-  initial = vernac_state;
-  of_sentence = SM.empty;
-}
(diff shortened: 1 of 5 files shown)


language-server/dm/executionManager.ml
```diff
@@ -51,13 +51,20 @@ type state = {
   initial : Vernacstate.t;
   of_sentence : (sentence_state * feedback_message list) SM.t;
   options : options;
+  doc_id : int; (* unique number used to interface with Coq's Feedback *)
```
maximedenes, line 54: Given the comment, it should be called `feedback_id` or something like that.

language-server/dm/executionManager.ml
```diff
@@ -51,13 +51,20 @@ type state = {
...
 }
 
-let init vernac_state = {
-  initial = vernac_state;
-  of_sentence = SM.empty;
-  options = default_options;
-}
+let init =
+  let doc_id = ref (-1) in
```
maximedenes, line 58: I found previous code (i.e. exposing a `fresh_xx_id` primitive) clearer than partial application with side effects, especially for OCaml non-experts.

language-server/dm/executionManager.ml
gares: A feedback message is `{ doc_id; route_id; span_id; contents }`. So to me `feedback_id` is even more confusing.

maximedenes (commented):
After discussion, `doc_id` should eventually disappear in favor of a global state in exec manager (which can then use globally unique `sentence_id` to multiplex feedback).

So we can merge this or not, as is most convenient.

PR #453, opened 2023-03-23
[vscoq2] opam release script

### LINKED ISSUE #401: automate release process for opam

gares:
Fix #401


gares:
The release script builds an opam repo which can be used to install the current language server and its coq branch.
The PR tests this in CI.

I hope it will be unnecessary soon, i.e. that we build against a released Coq version. Then, we could use dune-release (for example) to publish the language server, or opam publish if dune-release turns out to be too picky about the project layout.

PR #455, opened 2023-04-04
Fix problems with deleting lines

Hjaltesorgenfrei:
Deleting lines had (and still has after these changes) bugs with regards to state on the backend.
The first bug was that `interpret_to_end` got the last sentence id, and then did interpret to end. This meant that the sentence might have disappeared in the mean time and therefore that interpreting to that id was invalid.
The second bug was that when deleting a from sentence means it's `stop` was moved back and could mean that `stop < start`. To fix this I introduced `changed_sentences` which checks for all changed sentences. 
I don't know if this solution is the best as it checks on if the start and end changed. Checking if the length of the sentence changed might be a better idea. 

I also changed two todos which was in the code i was changing anyway.

Also I think this PR should be squashed when merging, we don't need the test print commits in the git history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ language-server/_build
 client/yarn-error.log
 client/.vscode-test/
 client/out/
+language-server/_opam/
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ language-server/_build
 client/yarn-error.log
 client/.vscode-test/
 client/out/
-language-server/_opam/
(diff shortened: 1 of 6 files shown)


maximedenes:
We analyzed the problem and this patch. Indeed there were a few bugs around sentence shift and deletion.

We went for an alternative fix: not shifting sentences on text edits, but reparsing the document.

So I'm closing this, but thanks for reporting and attempting a fix, it was very helpful.

PR #460, opened 2023-04-13
[VSCoq2] Speed up highlighting 

### LINKED ISSUE #454: [VsCoq 2] Slow parsing of files
As @Duckapple noted on zulip, parsing is currently very slow on large files. 
Testing it on Stdpp's List.v it takes 6 minutes to reach line 1711 on my machine, https://gitlab.mpi-sws.org/iris/stdpp/-/blob/master/stdpp/list.v#L1711
But if I delete everything after line 1725 and thereby make the file shorter it only takes 2 minutes to parse to the same point in the file. 
So there is probably some growth in parsing time of each sentence with regards to the total length of the file.

My theory is that `send_highlights` is the problem as it sends the highlight state for everything when something changes and not only for the changed part.

Hjaltesorgenfrei:
Fixes #454

This is faster due to not recreating the checked range each time highlights are sent.
By returning the ids from ExecutionManager.execute and keeping track of these it checks which have been executed.
Through I am not entirely happy with that implementation, returning an ID seems like an ugly design.
Maybe ExecutionManager and Document should each keep track of their respective parsed and checked ranges, which is then just called by DocumentManager?
What do you think @maximedenes 

Also I seem to have created a bug somewhere as the IDs from ExecutionManager.execute sometimes are not valid IDs in the document. Which is why remove_id_from_ranges and add_id_to_ranges are wrapped in a try statement.

### DIFF
--- a/language-server/dm/document.ml
+++ b/language-server/dm/document.ml
@@ -171,6 +171,7 @@ module ParsedDoc : sig
   val remove_sentences_after : t -> int -> t * Stateid.Set.t
   val sentences : t -> sentence list
   val sentences_before : t -> int -> sentence list
+  val surrounding_sentences : t -> sentence_id -> sentence option * sentence option
   val sentences_after : t -> int -> sentence list
   val get_sentence : t -> sentence_id -> sentence option
   val find_sentence : t -> int -> sentence option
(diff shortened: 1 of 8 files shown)


Hjaltesorgenfrei:
I fixed the problem, it's only one range now. But it also looked in the "checked" part and not legacy highlights.
Should we perhaps only have checked and parsed?

Hjaltesorgenfrei:
I see that you cache the highlighting in the client now. Should we close this PR?

maximedenes:
I think the bottleneck mainly comes from exchanging a lot of data to represent the highlights, which is fixed in #492. If another bottleneck happens to be in collecting the ids on the server side, we will revive this patch.

PR #476, opened 2023-05-23
Adding relevant screencaptures to the readme

rtetley:


rtetley:
@gares @maximedenes just let me know if you would like to ammend / change certain screen captures ?

maximedenes:
Maybe use key shortcuts for the one showing manual mode navigation?

gares:
I find the manual mode (step fwd/bwd) screencast to be polluted too much by the use of the "F1 + mnemonics".

IMO you could show once and for all that you have step fwd/bwd in there, and then use to the shortcuts to move the curtain back and forth, so to avoid the menu being unrolled every time, which is distracting.

PR #487, opened 2023-05-26
Remove syntactically invalid sentences from structured document view

### LINKED ISSUE #478: [Bug] Goal panel is buggy while editing text in continuous mode.
Expected behaviour: the goal panel should update as I type new tactics in continuous mode. 
Current behaviour: Tested on the first lemma of Asmgenproof0.v in backend from CompCert. When I type the tactics the goal panel behaves erratically, sometimes displaying "Not in proof mode." Or just not updating. 
Could be linked to #466

maximedenes:
Parsing errors are now handled specifically, and will be integrated into a concrete syntax view of the document when we have it.

Fixes #478.

### DIFF
--- a/language-server/dm/document.ml
+++ b/language-server/dm/document.ml
@@ -11,116 +11,23 @@
 (*   See LICENSE file.                                                    *)
 (*                                                                        *)
 (**************************************************************************)
-
 open Gramlib
+open Lsp
+open LspData
 open Types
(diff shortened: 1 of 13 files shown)


maximedenes:
This almost fixes #478, but there still a few corner cases to be addressed before merging this PR.

maximedenes:
This PR now fixes #478.

PR #494, opened 2023-06-05
Correct step navigation bugs

### LINKED ISSUE #493: Impossible to step backwards after first sentence
In manual mode, after doing "next" once and reaching the end of the first sentence, "back" has no effect.

rtetley:
Closes #493

### DIFF
--- a/language-server/dm/document.ml
+++ b/language-server/dm/document.ml
@@ -140,6 +140,9 @@ let find_sentence_after parsed loc =
   | Some (_, sentence) -> Some sentence
   | _ -> None
 
+let get_first_sentence parsed = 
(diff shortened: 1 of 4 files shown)


rtetley:
Corrected the step backward bug and another one that was overlooked: step forward when the document starts by empty space or comments did not work. 
@maximedenes I am not sure if reset is the right call when doing step backward at the first sentence of the document. Care to comment on that ?

maximedenes:
> I am not sure if reset is the right call when doing step backward at the first sentence of the document. Care to comment on that ?

As exchanged offline, I would prefer that we simply do a step back also in this case. Reset can be mapped to a dedicated command.

maximedenes (approved):

Issue #458, opened 2023-04-11
[VsCoq2] Context not correct for given location


Hjaltesorgenfrei:
`ExecutionManager.get_context` currently uses `Global.env ()` which means that auto completion suggests invalid lemmas which are defined later in the file. 

Example: 
```coq
Theorem proving_by_false : forall n m : nat, False -> n + m = m + n.
Proof.
    apply provin (* Using auto suggest here will suggest the lemma below. *)
Qed.


Theorem proving_by_false_again : forall n m : nat, False -> n + m = m + n.
Proof.
    intros. 
    exfalso.
    apply H.
Qed.
```

It is also not obvious how to fix this as the only other place i could see getting `Env` from is proofview, but that also uses the global context.

maximedenes:
I don't fully understand the problem described here: `ExecutionManager.get_context` indeed uses `Global.env ()`, but only after unfreeze the interp state at that point in the document (https://github.com/coq-community/vscoq/blob/ed40eda53d69dd17c14dce1542c83ddc81cab3e9/language-server/dm/executionManager.ml#L567).

PR #509, opened 2023-06-27
Merge improved autocomplete

### LINKED ISSUE #379: Inline suggestions for VSCoq 2

Duckapple:
Here is the type-aware code completion we (@Hjaltesorgenfrei, @Jakobis and I) have made for our Master's thesis [which can be read here](https://github.com/Jakobis/vscoqComparison/blob/main/Expanding_Coq_with_Type_Aware_Code_Completion.pdf).
While we implemented 5 algorithms for our thesis, through benchmarking we found out there is not that big of a reason for keeping all of them, as they perform very similarly, and the user wouldn't know which one to use anyways. We kept the two which work best for `apply` and `rewrite` respectively, such that we can potentially improve completion by switching to the better one of them depending on what is already written.
The benchmarking was done in this repository: https://github.com/Jakobis/vscoqComparison

Closes #379

### DIFF
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -584,17 +584,6 @@ let get_context st id =
       lemmas |> LemmaStack.with_top ~f:Proof.get_current_context |> Option.make
     end
 
-(*
-    match find_fulfilled_opt id st.of_sentence with
-    | Some (Success (Some { interp = ist; synterp = sst })) when Option.has_some ist.lemmas ->
-      Vernacstate.Interp.unfreeze_interp_state ist;
-      Vernacstate.Synterp.unfreeze sst;
-      let proof = get_proofview st pos in
-      None
-    | _ -> log "Cannot unfreeze state";
-    None   
-*)
-
 let get_completions st id =
   let aux () = 
     match find_fulfilled_opt id st.of_sentence with
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -531,6 +531,12 @@ let rec invalidate schedule id st =
   let deps = Scheduler.dependents schedule id in
   Stateid.Set.fold (invalidate schedule) deps { st with of_sentence }
 
+let unfreeze_interp_state st id =
+  match find_fulfilled_opt id st.of_sentence with
+  | Some (Success (Some { interp = st })) when Option.has_some st.lemmas ->
+    Vernacstate.Interp.unfreeze_interp_state st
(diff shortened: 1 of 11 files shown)


language-server/dm/executionManager.ml
```diff
@@ -569,6 +582,15 @@ let get_context st id =
     let sigma = Evd.from_env env in
     Some (sigma, env)
 
+let get_completions st pos =
+  unfreeze_interp_state st pos;
```
Hjaltesorgenfrei, line 586: Should the state be frozen again afterwards?

language-server/dm/executionManager.ml
```diff
@@ -569,6 +581,33 @@ let get_context st id =
...
+      None
+    | _ -> log "Cannot unfreeze state";
+    None   
+*)
+
+let get_completions st id =
+  let aux () = 
+    match find_fulfilled_opt id st.of_sentence with
+    | Some (Success (Some { interp = interp_state; synterp = synterp_state })) when Option.has_some interp_state.lemmas ->
+      Vernacstate.Interp.unfreeze_interp_state interp_state;
```
maximedenes, line 599: Why do you need to imperatively unfreeze the system state here?

language-server/dm/executionManager.ml
```diff
@@ -569,6 +581,33 @@ let get_context st id =
     let sigma = Evd.from_env env in
     Some (sigma, env)
 
+(*
+    match find_fulfilled_opt id st.of_sentence with
```
maximedenes, line 585: Leftover code

language-server/dm/executionManager.ml
Hjaltesorgenfrei: Because of this issue https://github.com/coq-community/vscoq/issues/458

maximedenes (approved):

PR #510, opened 2023-06-27
Enable find widget in goal view

Hjaltesorgenfrei:
An user requested the possibility to use the find widget in the goal view.
The change should not have any impact if users does not use ctrl+f


rtetley:
Thanks for this one !

PR #512, opened 2023-07-03
depend on sel 0.4.0

gares:
It finally works locally.
The main bug I introduced with the porting is that recurring events are no more available, one has to add the event back in the todo list by hand. Since I got it wrong, I was adding the lsp event twice and the code of SEL was not resilient to multiple events on the same file descriptor. I made the library resilient to that programming mistake in master.

The funny thing is that one event was reading even characters, the other odd characters, and both were hence failing to parse the content-length header.


gares:
The CI related commit is only partial, you get the log of the initialization phase but not the log that follows it.
I could not find the APIs to fetch and print the contents of the output window, so sad.

maximedenes:
Could we depend on sel rather than vendoring it?

From the description, it is not clear to me if the bugs you mention are fixed in sel 0.4.0.

gares:
If you fix nix we can stop vendoring it

gares:
To clarify, the bug was not there before the PR and is not introduced by the final version of this PR.

maximedenes:
The last two commits look unrelated to me, I'll move them to a separate PR.

maximedenes:
I open https://github.com/NixOS/nixpkgs/pull/247479 so that we can really unvendor sel.

gares:
> The last two commits look unrelated to me, I'll move them to a separate PR.

thanks!

PR #527, opened 2023-07-31
Add setting to enable/disable completion

maximedenes:
We disable it by default as it is still an experimental feature


maximedenes:
@rtetley does it look ok to you?

rtetley (approved):

Issue #524, opened 2023-07-13
Search Panel :  wrong semantic


thery:
when I search `bool and` it seems to search for theorems that talk about `bool` or `and`
while
```coq 
Search bool and.
```
search for theorems that talks about `bool` and `and`

PR #531, opened 2023-08-01
Search panel semantic problems

ThomasPortet:
Fixes the problem in #524, when using the search panel command with multiple patterns. The other problem was with restrictions. This PR adds the possibility to use the arguments in/inside/outside (see : (https://coq.inria.fr/refman/proof-engine/vernacular-commands.html?highlight=search#coq:cmd.Search) for the semantics).
For now the needed function interp_search_restriction is copied because it's not available in the coq API but once it is it should be deleted here and properly called.


rtetley:
Nice job ! Closes #524

maximedenes (approved):
Thanks!

PR #532, opened 2023-08-02
Sticky cursor for step forward and step backward

rtetley:


maximedenes (commented):
Maybe rename and redocument `last_executed_range` as it is really the range of observe id. Also it would be better to make sure that we 1. jump the cursor, 2. execute and 3. display the proof context.

maximedenes (approved):

PR #543, opened 2023-08-03
Diag channels

rtetley:
Appending the different types of coq feedbacks (Debug, Info, Notice) in separate channels instead of treating them as diagnostics

### DIFF
--- a/client/src/client.ts
+++ b/client/src/client.ts
@@ -3,14 +3,20 @@ import * as vscode from 'vscode';
 import {
   LanguageClient,
   LanguageClientOptions,
-  ServerOptions
+  ServerOptions,
+  integer
 } from 'vscode-languageclient/node';
 
 import {decorationsManual, decorationsContinuous} from './Decorations';
 
+import { FeedbackChannel } from './protocol/types';
+
 export default class Client extends LanguageClient {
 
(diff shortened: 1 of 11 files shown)


rtetley:
I am still not sure if this is the right way to go @maximedenes. 

- On the one hand, the lsp protocol specifies that the severity should be Error | Warning | Information | Hint: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticSeverity
- On the other hand, coq has a feedback level type which is: Debug | Info | Notice | Warning | Error: 
https://github.com/rtetley/coq/blob/ce3025d6b962b0c6e6fd109b229be615ca0e13a3/lib/feedback.ml#L13

How do we reconcile the two ? This PR aims at throwing Debug | Info | Notice into their own dedicated channels in the extension and Warning | Error in the usual Problems channel. 

Maybe at least the separation should be clearer in the backend. What do you think ?

maximedenes:
> Maybe at least the separation should be clearer in the backend. What do you think ?

Yes, I think the distinction should be clear: only Error and Warning should be emitted as diagnostics (for now). Notice & Info feedback should be collected by the server and sent for display to the dedicated channels. I'm not sure about Debug, but we can see that later.

maximedenes (approved):

PR #559, opened 2023-08-07
Syntax highlighting in proofview

### LINKED ISSUE #549: Restore syntax highlighting of proof view
We should, like VsCoq 1 does, highlight terms in the proofview.

maximedenes:
Fixes #549.


maximedenes:
@rtetley we'll have to refactor the goal-view code at some point, to better share type definitions and CSS classes.

PR #577, opened 2023-08-20
add instructions for beta release and Coq 8.18+rc1

palmskog:
As requested by @maximedenes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,44 @@ and contributors.
   compatible with Coq 8.18 or more recent, and supports manual or continuous mode
   checking.
 
+## Installing a VsCoq 2 beta release
+
+To use a beta release of VsCoq 2, you need to (1) install the VsCoq 2 language server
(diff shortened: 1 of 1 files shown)


maximedenes:
Thanks a lot for these improvements. Just one question: are you sure that pinning 8.18+rc1 is a good idea for end users? Won't it make the process less transparent for them once 8.18.0 is out?

palmskog:
The README obviously needs to be changed once 8.18.0 is out. These instructions are to make it easier for people to get started with testing/running betas.

palmskog:
Oh and specifically about the pinning: if you don't pin Coq to `8.18+rc1`, you will be prompted to upgrade to `coq.8.18.dev` now and then, which is probably not what anyone wants.

maximedenes:
>  you will be prompted to upgrade to coq.8.18.dev now and then

Ah! I see, thanks!

maximedenes (approved):

PR #615, opened 2023-09-12
Update screen capture and readme

rtetley:
Getting ready for vscoq 2 release I made a new set of screen captures and will update the readme.


rtetley:
@gares @maximedenes What do you think of this new set of screenshots for the release ?

gares:
The continuous checking gif is a bit odd(ly named).
You should that you can access any goal at any time, but you don't change the text and get continuous feedback on what you write. In some sense you could do the same screen capture in manual mode, don't you?

gares:
Also, the font in the query panel is a bit funky to me. In any case, it is different from the other fonts, so you ask yourself why, but that is not the point of the demo ;-)

rtetley:
> Also, the font in the query panel is a bit funky to me. In any case, it is different from the other fonts, so you ask yourself why, but that is not the point of the demo ;-)

Well that could be a mistake on my part... I'll check ! Thanks ;-)

rtetley:
> You show that you can access any goal at any time, but you don't change the text and get continuous feedback on what you write. In some sense you could do the same screen capture in manual mode, don't you?

Makes sense. I'll update it to show continuous typing / feedback. 
Also I am going to cleanup the gifs / gif names before merging !

rahulfromearth:
Hi @rtetley, What theme and font are you using in the gif? It looks aesthetic.

rtetley:
Hi @rahulfromearth, thanks ! The theme is Purrp!, and the font is Cartograph CF. Cheers !

PR #618, opened 2023-09-13
release 0.3.9

thery:


Zimmi48:
@thery Currently the release title / description is a bit strange (title : v0.3.9: Merge pull request #618 from thery/vscoq_0_3_9). Furthermore, this is marked as a pre-release, which I guess was not intended.

thery:
it was intended I did not want it the 2.0 to be the latest release

Zimmi48:
If what you mean is that you didn't want it to appear as the latest release, nowadays, GitHub has an option that you can untick for that:

image

It is already unticked, actually, so you can safely untick the "set as pre-release" one as well.

Furthermore, adding a title and a description would avoid it appearing like it does today:

image

thery:
@Zimmi48  Thanks. I've updated, hope it is ok now.

PR #629, opened 2023-09-19
Getting ready for the patch release

rtetley:
- Bump version
- Add readme instructions
- Correct 32 bit integer problem in opam CI

### DIFF
--- a/language-server/vscoqtop/lspManager.ml
+++ b/language-server/vscoqtop/lspManager.ml
@@ -40,7 +40,7 @@ let full_messages = ref false
 
 let Dm.Types.Log log = Dm.Log.mk_log "lspManager"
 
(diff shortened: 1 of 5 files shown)


language-server/vscoqtop/lspManager.ml
```diff
@@ -40,11 +40,11 @@ let full_messages = ref false
 
 let Dm.Types.Log log = Dm.Log.mk_log "lspManager"
 
-let conf_request_id = 3456736879
+let conf_request_id = max_int
```
gares, line 43: This will cause no compilation error on 32 bit ocaml, but the same source will result in different numbers depending on the architecture. I don't know if that could be a problem. Using `Int32` can avoid that.

PR #633, opened 2023-09-20
VsCoq downgrade message

4ever2:
Now that VsCoq 1 & 2 are different extensions on the marketplace it makes sense to point users directly to the installation page for VsCoq 1 in the error message.

Furthermore, it is now possible that both extension can be installed at the same time which is problematic. This PR attempts to detect if both extensions are enabled and warn the user that they should disable one of them.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -52,7 +52,7 @@ export function activate(context: ExtensionContext) {
             switch(err.status) {
 
                 case ToolChainErrorCode.notFound: 
-                    window.showErrorMessage(err.message, {title: "Install the VsCoq language server", id: 0}, {title: "Downgrade to VsCoq 0.3.9", id: 1})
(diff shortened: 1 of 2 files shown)


rtetley:
I have been working on something similar, except instead of pointing towards the page the action would install VsCoq Legacy and possibly disable VsCoq 2 (although I'm not sure I can do that with the vscode api). I also made the message a modal to make it clearer. 

Do you think it would be better to just point towards the marketplace page ? 

**
Screenshot 2023-09-25 at 09 33 01
**

rtetley:
I guess pointing towards the market place page seems like the most pragmatic choice after thinking about it. What do you think about the modal message ? Once we figure that out I'll propose a change and we can merge your work :-)

rtetley:
I've kept your modifications and just turned the message into a modal, I will merge once the check passes ! Thanks a lot for your contribution :-)

PR #646, opened 2023-10-02
Plug reset command

### LINKED ISSUE #606: Implement Reset command.
Add the reset command to VsCoq 2. 
Also, following #233, make sure that the vscoqtop process gets killed on reset.

rtetley:
Closes #606

### DIFF
--- a/client/goal-view-ui/src/App.tsx
+++ b/client/goal-view-ui/src/App.tsx
@@ -34,6 +34,10 @@ const app = () => {
                 }
             );
             break;
+        case 'reset':
+            setMessages([]);
+            setGoals(null);
+            break;
     }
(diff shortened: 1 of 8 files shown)


gares (approved):
lang server code looks OK

I hereby state my total disgust for using `'` (prime) in variable names, if only because it does not scale.

rtetley:
> lang server code looks OK
> 
> I hereby state my total disgust for using `'` (prime) in variable names, if only because it does not scale.

Haha okay ! For reference what would be the preferred denomination ?

gares:
I prefer `foo0` ... `foo9`, to `foo'` ... `foo'''''''''` ;-)

PR #650, opened 2023-10-03
Add a status bar item

### LINKED ISSUE #640: Viewing the VsCoq2 server executable and Coq version inside VS Code
Currently, the user can either specify the path to `vscoqtop`, or it can be read from path. In the latter case, since many people may have several opam switches, it's not obvious which executable was chosen, and if so, what the corresponding Coq version is.

Therefore, it would be useful to have the executable path and the Coq version (result of `vscoqtop --version`) displayed somewhere inside VS Code, e.g., in a panel or at least under extension settings. This is not only helpful for users themselves, but also when others are trying to help them debug some problem.

Another related problem, which deserves its own issue, is how to alternate between different opam switches/`vscoqtop`s from inside VS Code.

rtetley:
Add a status bar item to display vscoqtop info (version and coq version).
Closes #640

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -8,7 +8,8 @@ import {workspace, window, commands, ExtensionContext,
   Uri,
   StatusBarItem,
   extensions,
-  StatusBarAlignment
+  StatusBarAlignment,
+  MarkdownString
 } from 'vscode';
 
 import {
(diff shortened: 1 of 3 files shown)


rtetley:
@palmskog How do you feel about this for a display ? The large panel gets displayed on hover. Screenshot 2023-10-03 at 18 11 07

thery:
@rtetley Are you planning to do a release soon or should we use master?

rtetley:
@thery I was thinking of pushing for a release next week

rtetley:
@palmskog I will address the opam stuff in a separate issue but this could work well for it indeed

PR #656, opened 2023-10-10
Syntax error bug

### LINKED ISSUE #644: Can't get rid of Syntax Error: Lexer: Undefined token
Sometimes I can't get rid of a lexer problem. The expression is valid but it does not parse. 
Here is a way to reproduce it (not 100% sure it works)
I start with the code.
```coq
Variables ph pl : nat.
Lemma foo : (ph + pℓ − 0) <= 0.
```
it indicates an error on "0" now if I edit to remove the non ascii to get  (`ℓ −`) and get the correct code
```coq
Lemma foo : (ph + pl - 0) <= 0.
```
the lexer error remains

rtetley:
Closes #644.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -82,6 +82,8 @@
           self.packages.${system}.vscoq-language-server.buildInputs
           ++ (with ocamlPackages; [
             ocaml-lsp
+            ocamlformat-rpc-lib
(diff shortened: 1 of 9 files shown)


rtetley:
Closing in favour of #664

PR #664, opened 2023-10-12
Add support for multiple byte UTF-8 characters

tlsomers:
Positions in LSP are defined as positions in UTF-8 (or 16) characters, rather than number of bytes. This causes issues when converting from positions to locations and back when using longer symbols, such as `Φ`. In lines with such symbols, the last tactic could not be interpreted to correctly (when using the cursor).

This changes the conversion from location to position and position to location in order to keep track of the length of UTF-8 characters.


rtetley:
Thanks a lot for your contributions ! Closes #644.

PR #665, opened 2023-10-13
do not crash if a Coq sentence raises an anomaly

gares:
CC @maximedenes 

This code is there since the beginning, but I believe it is wrong.
If the exception reaches the toplevel, then the lang server dies.


gares:
I can't really make a test for this, since Coq has not command which raises an anomaly, but I tried it by hand and indeed it works.

rtetley:
@gares should I merge this or are we waiting for @maximedenes ?

PR #668, opened 2023-10-13
Remove auto extension publish

rtetley:
From now on the extension should be published manually, most of time only once the server has been published to opam.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -4,37 +4,6 @@ on:
     - published
 
 jobs:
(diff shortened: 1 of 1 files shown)


maximedenes:
Why not use a manual trigger, rather than removing the workflow?

rtetley:
I was under the impression that I did ? Maybe I forgot the file in the PR, I will check !

rtetley:
The manual trigger is in the the publish-extension.yml file. Sorry for the ambiguous title

PR #690, opened 2023-11-20
update docs and fix typo

tomtomjhj:


client/src/protocol/types.ts
```diff
@@ -51,7 +51,7 @@ export interface ProofViewNotification {
     messages: CoqMessage[];
 }
 
-export interface UpdateHightlightsNotification {
+export interface UpdateHighlightsNotification {
     uri: Uri; 
     parsedRange: Range[];
```
rtetley, line 56:
```suggestion
    processingRange: Range[];
```
After checking this is the current semantic. Thanks for the changes :-)

rtetley (commented):
Thanks for your work !

PR #693, opened 2023-11-23
Refactor observen id

rtetley:
This pull request adresses some inconsistencies with the observe id. 
By default now the observe id is None in continuous mode and Top (for top of the document) or (Id id) when on a specific sentence.


rtetley:
A problem is now changing the settings... 
If we change from Manual to Continuous, it's easy, we just pot observe_id = None
If we change form Continuous to Manual it is tricky, how to we assign observe_id ? Some ideas: we just assign it to top, or we look for the last sentence id which was executed and we assign it to that.

gares:
I would do the former, since computations are cached, we lose nothing: advancing forward will be instantaneous up to the point you would use in the latter option

rtetley:
Definitely simpler, however since decorations in manual mode depend upon the observe_id, the user will be under the impression that the execution state has been reset. Not great UX I guess.

gares:
If you are in continuons mode you should not see any green area, and hence going to manual is the same (there is no green disappearing).

Or you refer to other decorations?

rtetley:
You're right. I've changed the code so to reset the decorations when we move from manual to continuous mode as well !

PR #706, opened 2024-01-16
Updating vscode-test

rtetley:
This is due to a regression in the vscode-test repo. Temp fix until they fix their code

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -325,7 +325,7 @@
       }
     ],
     "menus": {
-       "editor/context": [
+      "editor/context": [
         {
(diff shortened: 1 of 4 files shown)


rtetley:
> This is due to a regression in the vscode-test repo. Temp fix until they fix their code
This has been fixed [here](https://github.com/microsoft/vscode-test/issues/255#event-11502837996). 
There is still an ongoing problem with coq platform ci, see https://github.com/coq/coq/issues/18548
Should be fixed by https://github.com/coq/coq/pull/18550

PR #715, opened 2024-01-26
Run `vscoq.path` as a shell command from the project directory

afdw:
This fixes an inconsistency between using `child_process.exec`, which runs the provided string as a shell command, and `vscode-languageclient`, which by default did not.

Also, `vscoqtop` is now always run from the directory of the project.

All that allows to use VsCoq for Coq projects built with dune, by putting the following to `.vscode/settings.json` (where `theories/Theory.v` is some Coq file):
```json
{
    "vscoq.path": "dune coq top --toplevel vscoqtop theories/Theory.v --"
}
```


rtetley:
Merging as CI issues have been resolved in main. Thanks  for your contribution !

PR #717, opened 2024-01-30
Nix flakes: Using the right ocamlPackages and updating locks

CohenCyril:
Builds on my machine

CC @rtetley

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -45,11 +45,11 @@
         "systems": "systems_2"
       },
       "locked": {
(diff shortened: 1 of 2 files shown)


CohenCyril:
btw it would be nice if you could reduce the nix deps to the bare minimum and inform me, or patch the derivation directly at https://github.com/NixOS/nixpkgs/pull/256515

rtetley:
Thanks a lot ! I'll reduce tomorrow and let you know ! Cheers :-)

PR #722, opened 2024-02-09
Adapt w.r.t. coq/coq#18652.

ppedrot:


gares (changes requested):
This change breaks 8.19 and 8.18.
We use ppx_optcomp that gives you ifdefs at the top level. There are a few examples around already, please follow that style

gares:
Example
https://github.com/coq-community/vscoq/blob/4eefe13822617d99de2e8bb5430f04209829effa/language-server/dm/parTactic.ml#L63

ppedrot:
Fixed.

gares:
@rtetley we should not fail fast in the matrix. Here coq master does not include the change yet, but we still want to test 8.18 and 8.19

gares (approved):

gares:
Also the opamfile ditches .dev, so we test coq 8.19 twice (the dev job downgrades it)

gares:
We should add a ... | = "dev" in the expression

gares:
Sorry for pushing here ci changes, but this is the first PR after the ppx_optcomp change, hence we need to streamline the process

.github/workflows/ci.yml
```diff
@@ -73,7 +74,7 @@ jobs:
         OPAMYES: true
       run: |
         opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev
-        opam install coq-core.${{ matrix.coq }}
+        opam pin add coq-core ${{ matrix.coq }}
```
gares, line 77:
```suggestion
        if [ "${{ matrix.coq }}" = "dev" ]; then opam install coq-core.dev; else opam pin add coq-core ${{ matrix.coq }}; fi
```

gares:
looks finally ok, the only jobs failing are the ones about coq master, since these are tested by Coq ci (against the PR requiring this change). The jobs about stable versions are green.

@rtetley the CI changes can be pushed to master, IMO. Dunno if we want the dev jobs to be in allowed-failure, or not. I did also spot a 8.18 value in the windows job, IIRC. Not sure it is actually used.

ppedrot:
There are unrelated changes in this PR, maybe it should still be marked as a draft while the upstream Coq PR is not merged...

PR #725, opened 2024-02-13
Fix ci

rtetley:
- Add a publish extension dry run to workflow
- Update publish extension action version.


rtetley:
Currently the CD pipeline is broken. Notified upstream here https://github.com/HaaLeo/publish-vscode-extension/issues/51

PR #726, opened 2024-02-13
Migrate to yarn v4

rtetley:


rtetley:
Turns out [vsce](https://github.com/microsoft/vscode-vsce) does not support yarn >= 2

PR #738, opened 2024-02-19
Fix coq-master CI

rtetley:


SkySkimmer:
Isn't it just that the docker image is outdated?

rtetley:
Actually you're probably right, it's the pin in the flake.nix that's outdated... Will fix in a minute

PR #739, opened 2024-02-20
Capture the correct loc for when an import fails.

rtetley:
Explicit the corresponding errors and get the loc from the qualid, which solves a silly display bug.


gares (changes requested):
please make a toplevel `get_loc_from_info_or_exn` factoring the code which is needed on 8.18, 8.19 but not master (on master it will be `Loc.get_loc info`)

rtetley:
Added the optcomp, and created https://github.com/coq/coq/pull/18694 correcting the coq API

rtetley:
@gares is this okay ?

gares (approved):

Issue #743, opened 2024-02-21
Provide a way to distinguish unparsed from successful


CohenCyril:
We need to distinguish
1. not yet parsed. (e.g. for long files and slow code)
2. parsed and successful
3. parsed and failed.

Right now 1., 2. print the same way, while 3. is underlined in red.

We could even add
4. parsing in progress

We could have color code (configurable by the user) to distinguish between 1. 2. and 4.
By default we could 
- highlighe in blue/green successful evaluations, (this would look like manual mode and be less destabilizing
- and/or overline in grey what is not yet processed at all (e.g. the end of the file)
- and optionally overline in orange things that are under evaluation, (so we can witness visually something is taking time).

PR #744, opened 2024-02-21
Better highlights

### LINKED ISSUE #642: Visual feedback for command in execution in manual mode
For tactic that takes some time, it would be nice to get a visual feed back of the command that is currently executed in Coq.

```coq
Parameter foo : nat -> Prop.
Axiom fooS : forall n, foo n -> foo n -> foo (S n).
Axiom fooO : foo 0.

Lemma foo14 : foo 14.
Time repeat apply fooS; apply fooO.
```

when executing the last line, having it in orange till it passes green would be nice

### LINKED ISSUE #657: Performance issue: Vscoq2 very slow on a file with numerous lemmas

rtetley:
WIP for making faster and better highlights
Adresses #743 for manual mode, need to think it through for continuous mode.
Closes #657 
Closes #642

### DIFF
--- a/language-server/dm/documentManager.ml
+++ b/language-server/dm/documentManager.ml
@@ -37,6 +37,7 @@ type state = {
   document : Document.document;
   execution_state : ExecutionManager.state;
   observe_id : observe_id option;
+  overview: exec_overview;
 }
 
 type event =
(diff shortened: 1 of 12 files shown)


rtetley:
This is preliminary work. It solves the performance issues for good (making them on par with VsCoq 1 on the linked example from #657. It also creates better distinctions between executed, executing and "prepared" (i.e. the range the user has asked after an interpret to point). I am not sure how well it works with delegation.
For continuous mode highlights I will handle it in a subsequent PR.
The general mechanism here will allow us to many QOL improvements down the line.
What do you think @gares ?

language-server/dm/documentManager.ml
```diff
@@ -291,7 +318,8 @@ let init init_vs ~opts uri ~text observe_id =
   let init_vs = Vernacstate.freeze_full_state () in
   let document = Document.create_document init_vs.Vernacstate.synterp text in
   let execution_state, feedback = ExecutionManager.init init_vs in
-  let st = { uri; opts; init_vs; document; execution_state; observe_id } in
+  let overview = { processing = []; processed = []; prepared = []} in
```
gares, line 321: nitpick: define an empty_overview since you do this twice

language-server/dm/executionManager.ml
```diff
@@ -220,6 +222,87 @@ let interp_qed_delayed ~proof_using ~state_id ~st =
...
+  if Position.compare r1.Range.end_ r.Range.start == 0 then
+    Range.{ start = r1.Range.start; end_ = r.Range.end_ } :: l
+  else
+    r1 :: (insert_or_merge_range r l)
+
+let rec remove_or_truncate_range r = function
+| [] -> []
+| r1 :: l ->
+  if Position.compare r1.Range.start r.Range.start == 0 && 
+    Position.compare r1.Range.end_ r.Range.end_ == 0
```
gares, line 237: this looks like Position.equal (if we derived one)

gares (approved):

PR #753, opened 2024-03-13
Remove document from state table when it is closed.

### LINKED ISSUE #751: Closed Files Still Use Memory
As mentioned on Zulip, when a file is closed in VSCoq2, the memory is still in use. (I am using VSCoq version 2.1.2) This means that after opening and interactively editing several files, `vscoqtop` uses huge amounts of memory (>6 GB for 2 files), and there is no way to reduce the memory usage except by closing the VSCode window. In VSCoq legacy, closing a file freed the memory associated with that file's Coq buffer.

rtetley:
Closes #751 

- [x] make the limit configurable
- [x] what to do when tabs are invisible but not purged?

### DIFF
--- a/language-server/vscoqtop/lspManager.ml
+++ b/language-server/vscoqtop/lspManager.ml
@@ -297,6 +297,9 @@ let consider_purge_invisible_tabs () =
   let usage = current_memory_usage () in
   if usage > 4000000000 (* 4G *) then begin
     purge_invisible_tabs ();
+    let vst, _ = get_init_state () in
+    Vernacstate.unfreeze_full_state vst;
+    Vernacstate.Interp.invalidate_cache ();
     Gc.compact ();
     let new_usage = current_memory_usage () in
     log @@ Printf.sprintf  "memory footprint %d -> %d" usage new_usage;
--- a/language-server/vscoqtop/lspManager.ml
+++ b/language-server/vscoqtop/lspManager.ml
@@ -29,7 +29,9 @@ let get_init_state () =
   | Some st -> st
   | None -> CErrors.anomaly Pp.(str "Initial state not available")
 
-let states : (string, Dm.DocumentManager.state) Hashtbl.t = Hashtbl.create 39
+type tab = { st : Dm.DocumentManager.state; visible : bool }
+
+let states : (string, tab) Hashtbl.t = Hashtbl.create 39
 
 let check_mode = ref Settings.Mode.Manual
 
@@ -212,23 +214,27 @@ let update_view uri st =
     publish_diagnostics uri st;
   )
 
+let replace_state path st visible = Hashtbl.replace states path { st; visible }
+
 let run_documents () =
-  let interpret_doc_in_bg path (st : Dm.DocumentManager.state) events =
-    let st = Dm.DocumentManager.clear_observe_id st in
-    let (st, events') = Dm.DocumentManager.interpret_in_background st in
-    let uri = DocumentUri.of_path path in
-    Hashtbl.replace states path st;
-    update_view uri st;
-    let events' = inject_dm_events (uri, events') in
-    events@events'
+  let interpret_doc_in_bg path { st : Dm.DocumentManager.state ; visible } events =
+    if not visible then events
+    else
+      let st = Dm.DocumentManager.clear_observe_id st in
+      let (st, events') = Dm.DocumentManager.interpret_in_background st in
+      let uri = DocumentUri.of_path path in
+      replace_state path st visible;
+      update_view uri st;
+      let events' = inject_dm_events (uri, events') in
+      events@events'
   in
   Hashtbl.fold interpret_doc_in_bg states []
 
 let reset_observe_ids =
-  let reset_doc_observe_id path (st : Dm.DocumentManager.state) events =
+  let reset_doc_observe_id path {st : Dm.DocumentManager.state; visible } events =
     let st = Dm.DocumentManager.reset_to_top st in
     let uri = DocumentUri.of_path path in
-    Hashtbl.replace states path st;
+    replace_state path st visible;
     update_view uri st
   in
   Hashtbl.fold reset_doc_observe_id states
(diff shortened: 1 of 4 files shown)


gares:
We could use ocaml GC module to know the memory footprint and let the user declare a cap above which we purge closed tabs.

gares:
@rtetley I did look at this PR implementing the memory reclaim when the memory footprint > 4G (should be configurable I guess).

There are two problems:
- I added a `visible` attribute to each tab, and there is a lot of code that is done on all tabs. Now we could only operate on visible tabs, but before this patch I really don't understand what the code was doing
- Removing tabs for that table does not seem to help much in reclaiming memory, so either I'm misusing the Gc API or the state is also referenced by something else. Try to open a large file, like Pfff.v and close the tab. I now print how much memory was reclaimed.

language-server/vscoqtop/lspManager.ml
```diff
@@ -211,23 +213,27 @@ let update_view uri st =
...
-  let interpret_doc_in_bg path (st : Dm.DocumentManager.state) events =
-    let st = Dm.DocumentManager.clear_observe_id st in
-    let (st, events') = Dm.DocumentManager.interpret_in_background st in
-    let uri = DocumentUri.of_path path in
-    Hashtbl.replace states path st;
-    update_view uri st;
-    let events' = inject_dm_events (uri, events') in
-    events@events'
+  let interpret_doc_in_bg path { st : Dm.DocumentManager.state ; visible } events =
+    if not visible then events
```
gares, line 220: This if makes me thing the current code does stuff on closed tabs, no?

language-server/vscoqtop/lspManager.ml
```diff
@@ -266,30 +272,59 @@ let textDocumentDidChange params =
...
+
 let textDocumentDidClose params =
-  [] (* TODO handle properly *)
+  let Lsp.Types.DidCloseTextDocumentParams.{ textDocument } = params in
+  let path = DocumentUri.to_path textDocument.uri in
+  match Hashtbl.find_opt states path with
+  | None -> assert false
+  | Some { st } ->
+      replace_state path st false;
+      consider_purge_invisible_tabs ();
```
SkySkimmer, line 311: maybe `st` is still seen as a GC root when purging? you could try
```suggestion
  begin match Hashtbl.find_opt states path with
  | None -> assert false
  | Some { st } ->
      replace_state path st false
  end;
  consider_purge_invisible_tabs ();
```

gares:
Thanks Gaetan, that clearly helped! Indeed the stack is a root. Kudos.

There is still something bad, I don't know if we should used some memory profiler.
If I open and close Pfff (and set the treshold to 0G, so that it always collects):
```
[lspManager          , 722507, 1719490340.485819] memory footprint 3967614976 -> 2680127488
```
while if I open a trivial file
```
[lspManager          , 722939, 1719490376.821685] memory footprint 93913088 -> 124223488
```
So clearly just parsing Pfff leaves a crapload of memory hanging

gares:
Closing the trivial buffer actually increses the memory, I don't know how Gc.compact can result in that.

gares:
*update* it is not reproducible, this time it went down to 500M, that is still huge, but 5 times smaller. So I guess the measure implemented in this PR is OKish.

The other question for @rtetley is still valid: why is the code doing something on invisible tabs? Is it wanted? Should I systematically skip working on invisible tabs (even if their state is not purged)?

SkySkimmer:
What is the state of the installed summary (eg the parser state, Global.env, etc) when running the Gc.compact?

gares:
Exactly, this is what could have changed, since this time I worked on the smaller tab before closing the huge one.

It means that calling compact without installing a minimal state may not work well. Looking into it.

gares:
Bingo! it now goes to 100M

language-server/vscoqtop/lspManager.ml
gares: Nope, it is OK to do the things on invisible docs. For example if a settings change.

language-server/vscoqtop/lspManager.ml
```diff
@@ -211,23 +213,27 @@ let update_view uri st =
...
-    let st = Dm.DocumentManager.clear_observe_id st in
-    let (st, events') = Dm.DocumentManager.interpret_in_background st in
-    let uri = DocumentUri.of_path path in
-    Hashtbl.replace states path st;
-    update_view uri st;
-    let events' = inject_dm_events (uri, events') in
-    events@events'
+  let interpret_doc_in_bg path { st : Dm.DocumentManager.state ; visible } events =
+    if not visible then events
+    else
```
gares, line 221:
```suggestion
```

language-server/vscoqtop/lspManager.ml
```diff
@@ -266,30 +270,62 @@ let textDocumentDidChange params =
...
+    let new_usage = current_memory_usage () in
+    log @@ Printf.sprintf  "memory footprint %d -> %d" usage new_usage;
+  end
+
 let textDocumentDidClose params =
+  let Lsp.Types.DidCloseTextDocumentParams.{ textDocument } = params in
+  let path = DocumentUri.to_path textDocument.uri in
+  begin match Hashtbl.find_opt states path with
+  | None -> assert false
+  | Some { st } -> replace_state path st false
```
SkySkimmer, line 310: What's the reason for doing `replace` with `visible = false` then filtering instead of doing `Hashtbl.remove`?

rtetley: You don't necessarily want to remove the execution state of the document when you close it. For example you might close a document by mistake.

gares: The point is that, if memory allows for it, we want to keep the coq sessions alive.

It is not uncommon to close a tab and reopen it.
In VSCode you have have the usual browser shortcut for it, ctrl-w to close and shift-ctrl-t to open the last closed tab.
In this scenario we want to be nice to the users and not lose their session.

SkySkimmer: Does reopening the tab get the saved state with the current code?

Also I guess the consider_purge should be delayed a bit since big sessions are the ones that are the worst to lose by mistake.

rtetley: Good question. Probably not at the moment.

For the consider_purge a reasonable alternative would be to send a popup message to the user when memory runs out asking him if he wants to purge ?

rtetley:
I think we are good to go. @gares ?

PR #760, opened 2024-04-01
Improve flake.nix, add compatibility with pkgs.vscode extensions field

redanaheim:
This pull request adds two new outputs, packages.${system}.vscoq-client.[extension, vsix_archive].

The `extension` derivation builds the out folder in such a way that this derivation can be passed to vscode in Nix to automatically load it as an extension. Specifically, the source (including webpack results in the dist folder) are placed in $out/share/vscode/extensions/maximedenes.vscoq.
This is very useful for NixOS users or even people who use a shell.nix to configure their development environment.

The `vsix_archive` derivation builds the .vsix and places it in $out/share/vscode/extensions.

This PR also adds the lightweight Alejandra formatter output in flake.nix, which is automatically picked up by nix when running the command `nix fmt`. This is useful for two reasons:
1. It helps keep the code style consistent in the long flake.nix file.
2. It lets us use `nix fmt` in CI, which avoids failing CI when whitespace or other insignificant changes occur in the generated yarn.nix files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,9 +45,13 @@ jobs:
         nix_path: nixpkgs=channel:nixos-unstable
     - name: Check each yarn.nix up to date
       run: |
-        cd client && nix-shell -p yarn2nix --run "yarn2nix > new_yarn.nix" && diff new_yarn.nix yarn.nix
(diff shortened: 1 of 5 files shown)


redanaheim:
Not sure why the one CI check that doesn't even touch Nix failed but I'll look at it

rtetley:
Thanks for your contribution ! ;-)

PR #767, opened 2024-05-06
[coq] Adapt to coq/coq#18890

ejgallego:


herbelin:
I merged coq/coq#18890 without realizing that it had a vscoq overlay. To me merged when possible.

PR #771, opened 2024-05-19
Add Equations and Equations? to coq.tmLanguage.json to get coloring

thomas-lamiaux:
Add Equations and Equations? to coq.tmLanguage.json to get coloring


rtetley:
Thanks !

PR #773, opened 2024-05-30
Formatting proof view

### LINKED ISSUE #749: Formatting Proof view
Legacy vscoq allows proofs to be formatted nicely, But this is no the case for the current vscoq as shown below:

image

This is makes it extremely hard to follow computer science related proofs as the induction principles generated are very large and wordy. I often find myself having to copy paste it to a new tab and reformatting it by hand just to understand what I am trying to do.

rtetley:
Closes #749


client/goal-view-ui/src/utilities/pp-break.tsx
```diff
@@ -0,0 +1,57 @@
...
+                <span id={id}>
+                    <br/>
+                    <span style={style}>
+                        {" ".repeat(indent ? indent : 0)}
+                    </span>
+                </span>
+            );
+        case PpMode.hvBox:
+            if(lineBreak) {
+                <span id={id}>
```
tomtomjhj, line 34: missing `return` here?

rtetley: Good catch ! I think handling hvBoxes correctly stilll needs a bit of work anyways !

PR #782, opened 2024-06-07
Adding a warning as errors job in CI

rtetley:


gares:
looks good to me, but the last commit shall be removed

PR #795, opened 2024-06-24
Capabilities for outline

rtetley:


rtetley:
@gares if you can briefly review this then I'll merge and use it as a basis for further work using the "outline"

gares (commented):
My only question is about invalidation. What do you do?
It seems you do nothing when a sentence is removed and I guess you should follow what the scheduler does.

gares:
A nitpick. In OCaml `let... in` is well delimited, so if in the body you use a `match`, you don't necessarily need a `begin` `end`.

language-server/dm/document.ml
```diff
@@ -111,10 +179,13 @@ let add_sentence parsed parsing_start start stop (ast: parsed_ast) synterp_state
...
+    parsed with sentences_by_end = LM.add stop sentence parsed.sentences_by_end;
     sentences_by_id = SM.add id sentence parsed.sentences_by_id;
-    schedule
-  }, scheduler_state_after
+    schedule;
+  } in
+  let outline = record_outline document id ast.ast ast.classification parsed.outline in
+  {document with outline}, scheduler_state_after
 
 let remove_sentence parsed id =
```
rtetley, line 190: > My only question is about invalidation. What do you do? It seems you do nothing when a sentence is removed and I guess you should follow what the scheduler does.

@gares you mean what the scheduler does here ? :-P

PR #801, opened 2024-06-26
Fix errors when recovering from interp-error

### LINKED ISSUE #634: [vscoq2] language server hangs up with some specific erroneous inputs
Hi all. First, thank you all for your active development of this fantastic extension!

I tried using the extension to solve exercises of [Software Foundations](https://softwarefoundations.cis.upenn.edu), and I found the language server keeps failing and finally hangs up when I edit the source code. Here is a minified source code that reproduces this behavior.

```coq
Inductive bool : Type :=
  | true
  | false.

Definition negb (b:bool) : bool :=
  match b with
  | true => false
  | false => true
  end.

Definition andb3 (b1:bool) (b2:bool) (b3:bool) : bool :=
  if b1 then
    if b2 then
      if b3 then
        true
      else
        false
    else
      false
  else
    false.

Example test_andb33:                 (andb3 true false true) = false.
(* Comment from reporter: Admitted is missing here*)

Example test_andb34:                 (andb3 true true false) = false.
Admitted.

Module NatPlayground2.

Fixpoint plus (n : nat) (m : nat) : nat :=
  match n with
  | O => m
  | S n' => S (plus n' m)
  end.

Compute (plus 3 2).
(* Comment from reporter: Missing `End NatPlayground2.`*)
```

And I use the following environments:
* OS: WSL Ubuntu 22.04
* coq: 8.18.0
* vscode-langauge-server: 2.0.1 (installed via git repo)

Feel free to let me know if you need further information.

rtetley:
Just add an error guard, this prevents the full server from crashing but since I only send the original state could lead to some undesired effects.
See [here](https://coq.zulipchat.com/#narrow/stream/237662-VsCoq-devs-.26-users/topic/coq.20language.20server.20repeated.20crashes).

Closes #634

### DIFF
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -168,11 +168,9 @@ let interp_error_recovery strategy st : Vernacstate.t =
           Vernacstate.Declare.set (lemmas,program) [@ocaml.warning "-3"];
           let interp = Vernacstate.Interp.freeze_interp_state () in
           { st with interp }
-        | Error(_,_) -> 
-          (* let loc = Loc.get_loc info in
-          let msg = CErrors.iprint (e, info) in
-          let status = error loc msg st in
-          let st = interp_error_recovery error_recovery st in *)
+        | Error (Sys.Break, _ as exn) ->
+          Exninfo.iraise exn
+        | Error(_,_) ->
           st
 
 (* just a wrapper around vernac interp *)
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -155,12 +155,25 @@ let interp_error_recovery strategy st : Vernacstate.t =
     | Some lemmas ->
         let proof = Vernacstate.LemmaStack.get_top lemmas in
         let pm = NeList.head program in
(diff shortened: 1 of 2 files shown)


rtetley:
Sometimes there were uncaught exceptions in the interp_error_recovery function. If that happens I don't really know what to do about the state. Is this satisfying @gares ?

language-server/dm/executionManager.ml
```diff
@@ -155,12 +155,25 @@ let interp_error_recovery strategy st : Vernacstate.t =
...
+          Error (e, info)
+        in
+        match result with
+        | Ok (pm) ->
+          let lemmas = snd (Vernacstate.LemmaStack.pop lemmas) in
+          let program = NeList.map_head (fun _ -> pm) program in
+          Vernacstate.Declare.set (lemmas,program) [@ocaml.warning "-3"];
+          let interp = Vernacstate.Interp.freeze_interp_state () in
+          { st with interp }
+        | Error(_,_) ->
```
gares, line 171: please add a case like below for Sys.Break

PR #805, opened 2024-06-27
Fix highlights

### LINKED ISSUE #803: vscoq 2.1.3, break seems broken
Here is my example
```coq
Goal forall A B C D E F : Prop, A -> B -> C -> D -> E -> F -> A.
Proof.
intros A B C D E F.
intros HA; intros HB; intros HC; intros HD; intros HE; intros HF; exact HA.
Qed.
```
if I execute everything
ddd1
if I try to break the line after `HE;`
The interface goes into an inconsistent state
ddd2

rtetley:
When there is a non-invalidating edit, the exec overview needs to be shifted (like for the diagnostics).
Closes #803


language-server/dm/document.ml
```diff
@@ -144,7 +144,7 @@ let compare_code_line x y =
 
 let code_lines_sorted_by_loc parsed =
   List.sort compare_code_line @@ List.concat [
-    (List.map (fun (_,x) -> Sentence x) @@ SM.bindings parsed.sentences_by_id) ;
+    (List.map (fun (_,x) -> Sentence x) @@ LM.bindings parsed.sentences_by_end) ;
```
gares, line 147: this should be reverted, or better the two maps should be printed for future debugging

language-server/dm/document.ml
```diff
@@ -226,16 +226,24 @@ let pos_at_end parsed =
...
     Scheduler.schedule_sentence (id,ast) scheduler_state_before parsed.schedule
   in
   let new_sentence = { old_sentence with ast; parsing_start; start; stop; scheduler_state_before; scheduler_state_after } in
   let sentences_by_id = SM.add id new_sentence parsed.sentences_by_id in
-  let sentences_by_end = LM.remove old_sentence.stop parsed.sentences_by_end in
+  let sentences_by_end = match LM.find_opt old_sentence.stop parsed.sentences_by_end with
+  | Some { id } when Stateid.equal id new_sentence.id ->
+    LM.remove old_sentence.stop parsed.sentences_by_end 
+  | _ -> parsed.sentences_by_end
+  in
```
gares, line 246: CC @maximedenes

gares:
It would be nice to have some unit tests on the contents of `LM` and `SM` after document validation, although the bug we fix here is really hard to imagine in a test

gares:
If we remove the debug prints I believe we could merge and release

PR #811, opened 2024-06-28
[vscoq1] fix https://github.com/coq/coq/pull/19040

thery:


thery:
ok it seems to work

thery:
@herbelin @jfehrle  I am interesting in an example that shows something that was  wrong and has been fixed.
Thanks

herbelin:
I guess anything with an error should be noticeable, like `Check 1 1.`. You can also try if errors spanning over several lines are well-located.

I could make some tests too if you tell me how to test vscoq1 from a local archive.

thery:
If you enable the extension VsCoq Legacy when using vscode, you should get the interface running

This following test seems ok 

ddd1

Zimmi48:
Shouldn't there be a release or tag so that people find about the new version? Besides, the new version is only available on the Microsoft marketplace, not on Open VSX. Who is in charge of publishing new versions there?

thery:
my bad! I did the [tag](https://github.com/coq-community/vscoq/releases/tag/v0.4.1). Unfornately it seems that I have not the right to publish it since it is under `coq-community`.
@rtetley can you do it or give me the right

rtetley:
@thery just added you as a namespace owner !

thery:
@rtetley Thanks!

The package has been published

PR #828, opened 2024-07-16
Fix highlight bug

### LINKED ISSUE #810: Confusing behavior of "executed range"
With the following file as playground.

```
Require Import List Reals.
Open Scope R_scope.

Definition fib_aux :=
  nat_rect (fun _ => list R) (0 :: 1 :: nil)
    (fun _ v => (nth 1 v 0 :: nth 0 v 0 + nth 1 v 0 :: nil))
.

Definition fib (n : nat) : R :=
  nth 0 (fib_aux n) 0.

Lemma big_fib1 : fib 10 = 55.
Proof.
unfold fib; simpl; ring.
Qed.

Lemma big_fib : fib 11 = 89.
Proof.
unfold fib; simpl.
ring_simplify.
```
With Vscoq>Proof:Mode set to settings.
When I do alt-rightarrow  to line 14, then alt-downarrow, then 4 times alt-uparrow, I get the "executed zone" colored down to line 15 (should be to line 11 or 12).

Here is an animation showing this unwanted behavior https://filesender.renater.fr/?s=download&token=d76d35d8-2abf-48cd-9593-4c0ca0b4574a

rtetley:
Closes #810.

### DIFF
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -78,9 +78,8 @@ type state = {
   overview: exec_overview;
 }
 
-
-let print_exec_overview st =
-  let {processing; processed; prepared} = st.overview in
+let print_exec_overview overview =
+  let {processing; processed; prepared } = overview in
   log @@ "--------- Prepared ranges ---------";
   List.iter (fun r -> log @@ Range.to_string r) prepared;
   log @@ "-------------------------------------";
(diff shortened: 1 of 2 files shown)


rtetley:
While fixing this, I noted that the Range of a ```Qed.``` encapsulates the entire proof block (i.e. ```Proof. [...]. Qed.```).
Is this expected @gares ?

gares:
No, it is not

rtetley:
We should investigate

PR #829, opened 2024-07-16
Stop recomputing display for big goals

### LINKED ISSUE #809: vscoq2.1.3 big output breaks display of goal
The following code displays correctly for small values of fib (up to 11) and breaks starting at (12).  What breaks is that the goal window is empty.  Normal behavoir can only by restored by entering "Developer:Reload Window".

```
Require Import List Reals.
Open Scope R_scope.

Definition fib_aux :=
  nat_rect (fun _ => list R) (0 :: 1 :: nil)
    (fun _ v => (nth 1 v 0 :: nth 0 v 0 + nth 1 v 0 :: nil))
.

Definition fib (n : nat) : R :=
  nth 0 (fib_aux n) 0.

Lemma big_fib : fib 12 = 10.
Proof.
unfold fib; simpl.
```

rtetley:
Closes #809

### DIFF
--- a/client/goal-view-ui/src/utilities/pp.tsx
+++ b/client/goal-view-ui/src/utilities/pp.tsx
@@ -11,6 +11,8 @@ import classes from './Pp.module.css';
 import { off } from 'process';
 
 
+const MAX_RECOMPUTE = 50;
+
 type PpProps = {
     pp: PpString;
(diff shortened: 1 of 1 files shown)


rtetley:
It would be nice to find a way to increase the "maximum depth" in react, right now it seems to be set at 50

PR #833, opened 2024-07-16
allows newer ppx_yojson_conf version

CodiePP:
in newer ppx_yojson_conv versions apparently for deriving yojson types of primitives one has to include some declarations.
I was using v0.17 and this solved the compilation errors.

the opam file had a duplicate entry for the ppx_yojson_conv library. 🍤

### DIFF
--- a/language-server/vscoq-language-server.opam
+++ b/language-server/vscoq-language-server.opam
@@ -21,7 +21,6 @@ depends: [
   "ppx_inline_test"
   "ppx_assert"
   "ppx_sexp_conv"
-  "ppx_yojson_conv" {< "v0.16.0"}
   "ppx_deriving"
   "sexplib"
   "ppx_yojson_conv"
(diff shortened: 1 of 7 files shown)


gares:
There was an upper bound, dunno why, but I'd be happy to remove it. Unfortunately Ci is not picking the version you mention up, maybe because ocaml is 4.13? We can surely switch to 4.14 in the opam job. Would you mind trying?

language-server/vscoq-language-server.opam
```diff
@@ -11,7 +11,7 @@ build: [
   [ "dune" "build" "-p" name "-j" jobs ]
 ]
 depends: [
-  "ocaml" { >= "4.13.1" }
+  "ocaml" { >= "4.14" }
```
rtetley, line 14: Why the new lower bound ? Do your changes not work with ocaml < 4.14 ?

CodiePP: yes, it requires at least 4.14.0 (see https://ocaml.org/p/ppx_yojson_conv/v0.16.0)
and, now it is more consistent in the repo among flake.nix and ci, cd Github actions

rtetley: @gares I think that sounds find ! You can make the final call to merge !

PR #841, opened 2024-07-22
Don't error on deprecated warnings in ocaml code

SkySkimmer:


ppedrot:
CI green, is this reasonable a change?

gares:
I thought the dev provide has `-warn-error -all` (assuming this is the right syntax to tell dune that warnings are warnings).
So this PR is fine, but it could go one step furhter.

language-server/dune
```diff
@@ -3,9 +3,9 @@
     (deps vscoq-language-server.install))
 (env
  (dev
-  (flags :standard -w -9))
+  (flags :standard -w -9 -warn-error -3))
```
gares, line 6:
```suggestion
  (flags :standard -w -9 -warn-error -A))
```

gares:
still, if you are about to add a warning in master, then the job `coq-master, fatalwarnings` is going to fail anyway.
I can help fixing the usages of the deprecated symbol

SkySkimmer:
dev profile is warn-error +all not -all

gares:
I'm not sure I get it. In the file edited by this PR we redefine the dev profile and we add the fatalwarnings one.
In the dev profile we want warnings to be warnings.
In the fatalwarnings profile we want warnings to be errors.

Did we screw up @SkySkimmer ?

SkySkimmer:
> In the dev profile we want warnings to be warnings.

That is the opposite of the dune default but you can do that if you want.

Note that coq CI currently uses the dev profile.

gares:
> That is the opposite of the dune default but you can do that if you want.

I know, they are wrong.

> Note that coq CI currently uses the dev profile.

Looks reasonable to me. Maybe the coq CI should use a build system independent way to force warning as errors, and we should honor that in our makefile. Basing on the dune profile cannot cut it (some projects don't even use dune).

SkySkimmer:
I don't think deprecation warnings should be errors in coq CI, the point of deprecating something is to let the 3rd party developers write the adaptation in their own time instead of blocking the coq PR.

SkySkimmer:
>Maybe the coq CI should use a build system independent way to force warning as errors

that sounds impossible to me

SkySkimmer:
Shall we merge?

PR #871, opened 2024-08-20
feat: adding alt + shift click to expand all ellipsis goals

Durbatuluk1701:
Given the following case where you have a goal that is collapsed:
image

You can use `alt + click` to expand it to something like this:
image

However, I frequently find myself wanting to expand all the way if I am doing expanding, so I added a feature such that `alt + shift + click` will expand the original to this instead (note that no ellipsis remain as it will expand everything below the clicked on ellipsis):
image

Additionally, I found some typos while adding this feature and fixing them is included as well.

### DIFF
--- a/client/pp-display/src/pp-box.tsx
+++ b/client/pp-display/src/pp-box.tsx
@@ -1,13 +1,25 @@
 import React, {FunctionComponent, useEffect, useState, useLayoutEffect, useRef, ReactFragment, SyntheticEvent} from 'react';
-import {Box, DisplayType, BreakInfo} from './types';
+import {Box, DisplayType, BreakInfo, HideStates } from './types';
 import PpBreak from './pp-break';
 import classes from './Pp.module.css';
+
+/**
+ * Computes the optimal hide state given a hide state for the current box and the hide state for the parent box .
+ */
+const ComputeHideState = (self : HideStates, parent : HideStates) => {
+  // If "parent" state dictates "self", then follow it
+  if (parent === HideStates.HIDE || parent === HideStates.EXPAND_ALL) {
+    return parent;
+  }
+  // Otherwise, follow the self state
+  return self;
+}
 
 interface PpBoxProps extends Box {
     coqCss: CSSModuleClasses,
     breaks: BreakInfo[],
     maxDepth: number,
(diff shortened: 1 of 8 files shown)


client/pp-display/src/pp-box.tsx
```diff
@@ -16,8 +28,8 @@ const ADDED_DEPTH_FACTOR = 10;
 
 const PpBox: FunctionComponent<PpBoxProps> = (props) => {
     
-    const {mode, depth, coqCss, id, indent, breaks, boxChildren, hovered, maxDepth, addedDepth} = props;
-    const [hide, setHide] = useState<boolean>(depth >= maxDepth);
+    const {mode, depth, coqCss, id, indent, breaks, boxChildren, parentHide, hovered, maxDepth, addedDepth} = props;
+    const [selfHide, setSelfHide] = useState<HideStates>(ComputeHideState(depth >= maxDepth ? HideStates.HIDE : HideStates.UNHIDE, parentHide));
     const [depthOpen, setDepthOpen] = useState<number>(addedDepth);
```
Durbatuluk1701, line 33: One thing I might add, I was a bit confused about the `depthOpen` variables usage, so maybe there is a simpler solution to the "expand all the way" problem using that variable?

TheoWinterhalter:
I wonder, should it not be a button "Unfold all"? This way it's more accessible to newcomers?

client/pp-display/src/pp-box.tsx
rtetley: ```depthOpen``` controls at what depth an ellipsis will open. So when you click on an ellipsis, instead of only unfolding it's child, it will go down ```depthOpen``` which is more comfortable than having to unfold one by one.

rtetley: Sorry to clarity, ```depthOpen``` tracks the current depth at which an ellipsis happens, so the depth until which eveything is open. The ```ADDED_DEPTH_FACTOR``` just allows us to increment this in steps larger than 1.

rtetley:
Thanks for your contribution ! Just took the liberty of documenting the shortcuts. Concerning the Open all button we can move that to a separate issue !

PR #874, opened 2024-08-24
Changing feedback messages to be appended

### LINKED ISSUE #873: Debug output is printed reverse
Hi, I was trying debug auto with VsCoq2 and it seems that messages are printed in the reverse order. The following returns

```Coq
Goal forall {P Q R}, (P -> Q) -> R -> P -> Q.
  intros. debug auto.

** assumption. (*success*)
* simple apply X. (*success*)
* intro. (*fail*)
* assumption. (*fail*)
(* debug auto: *)
```
which seems reversed to me

Durbatuluk1701:
This PR changes the behavior of handlings feedback messages to append the message rather than prepend them.
 
This should specifically fix #873

### DIFF
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -530,7 +530,7 @@ let handle_feedback id fb state =
   match fb with
   | (_, _, _, msg) ->
     begin match SM.find id state.of_sentence with
-    | (s,fl) -> update_all id s (fb::fl) state
+    | (s,fl) -> update_all id s (fl @ [fb]) state
     | exception Not_found -> 
         log @@ "Received feedback on non-existing state id " ^ Stateid.to_string id ^ ": " ^ Pp.string_of_ppcmds msg;
         state


Durbatuluk1701:
One thing I also noticed while debugging this was that it seems to me like `debug auto` gets broken up into many different messages and events. Maybe at some point we could explore a more efficient solution of processing one large event/message since all the currently generated events/messages apply to the same range and have the same severity?

rtetley:
Thanks !

rtetley:
> One thing I also noticed while debugging this was that it seems to me like `debug auto` gets broken up into many different messages and events. Maybe at some point we could explore a more efficient solution of processing one large event/message since all the currently generated events/messages apply to the same range and have the same severity?

I agree this could be a good idea. Could you write it up in an issue to track this ?

gares:
I suspect that for long output, like auto traces, it is better to still prepend but rev when displaying (less allocations, better complexity)

Durbatuluk1701:
> I suspect that for long output, like auto traces, it is better to still prepend but rev when displaying (less allocations, better complexity)

I was a bit worried that if any other tools ever tried to interact with vscoqtop they might be confused unless they also reversed feedback messages client side. 
If you think it is worth it though I can definitely try the prepend and rev client-side approach.
Also might be worth considering if the append price is too high if #878 works out.

PR #875, opened 2024-08-25
feat: adding option for interpreting to exact cursor position or next command during interpret to point

### LINKED ISSUE #870: improvement for vim navigation
Using vscoq's interpret to point command, the current line of proof that the cursor is on is not interpreted. This causes a bit of inconvenience when using with the vscode vim extension. Basically using vim, it is impossible to place the cursor in such a way that the interpret to point command interprets the current line. This is inconsistent with the behavior of proof general in emacs where the interpret to point includes the line of proof that the cursor is currently on.

Durbatuluk1701:
This PR adds a "interpret to next point" that will take the current cursor position and if it is part of a tactic then it will interpret through the completion of that tactic.

This should fix #870 

Some test cases have also been added to test the function to find the end of the next tactic.

I am not sure if we want this to be the default behavior for continuous mode or should be an option that can be enabled.

### DIFF
--- a/language-server/protocol/extProtocol.ml
+++ b/language-server/protocol/extProtocol.ml
@@ -26,15 +26,7 @@ module Notification = struct
       type t = {
         textDocument : VersionedTextDocumentIdentifier.t;
         position : Position.t;
-      } [@@deriving yojson]
-
-    end
-
-    module InterpretToNextPointParams = struct
-
-      type t = {
-        textDocument : VersionedTextDocumentIdentifier.t;
-        position : Position.t;
+        to_next_point : bool
       } [@@deriving yojson]
 
     end
@@ -69,7 +61,6 @@ module Notification = struct
     | Std of Lsp.Client_notification.t
     | InterpretToEnd of InterpretToEndParams.t
     | InterpretToPoint of InterpretToPointParams.t
-    | InterpretToNextPoint of InterpretToNextPointParams.t
     | StepForward of StepForwardParams.t
     | StepBackward of StepBackwardParams.t
 
@@ -79,9 +70,6 @@ module Notification = struct
       | "vscoq/interpretToPoint" ->
         let+ params = Lsp.Import.Json.message_params params InterpretToPointParams.t_of_yojson in
         InterpretToPoint params
-      | "vscoq/interpretToNextPoint" ->
-        let+ params = Lsp.Import.Json.message_params params InterpretToNextPointParams.t_of_yojson in
-        InterpretToNextPoint params
       | "vscoq/stepBackward" ->
         let+ params = Lsp.Import.Json.message_params params StepBackwardParams.t_of_yojson in
         StepBackward params
--- a/language-server/protocol/extProtocol.ml
+++ b/language-server/protocol/extProtocol.ml
@@ -30,6 +30,15 @@ module Notification = struct
 
     end
 
+    module InterpretToNextPointParams = struct
+
+      type t = {
+        textDocument : VersionedTextDocumentIdentifier.t;
+        position : Position.t;
+      } [@@deriving yojson]
+
+    end
+
     module InterpretToEndParams = struct
 
       type t = {
@@ -60,6 +69,7 @@ module Notification = struct
     | Std of Lsp.Client_notification.t
     | InterpretToEnd of InterpretToEndParams.t
     | InterpretToPoint of InterpretToPointParams.t
+    | InterpretToNextPoint of InterpretToNextPointParams.t
     | StepForward of StepForwardParams.t
     | StepBackward of StepBackwardParams.t
 
@@ -69,6 +79,9 @@ module Notification = struct
       | "vscoq/interpretToPoint" ->
         let+ params = Lsp.Import.Json.message_params params InterpretToPointParams.t_of_yojson in
         InterpretToPoint params
+      | "vscoq/interpretToNextPoint" ->
+        let+ params = Lsp.Import.Json.message_params params InterpretToNextPointParams.t_of_yojson in
+        InterpretToNextPoint params
       | "vscoq/stepBackward" ->
         let+ params = Lsp.Import.Json.message_params params StepBackwardParams.t_of_yojson in
         StepBackward params
--- a/language-server/protocol/extProtocol.ml
+++ b/language-server/protocol/extProtocol.ml
@@ -26,7 +26,6 @@ module Notification = struct
       type t = {
         textDocument : VersionedTextDocumentIdentifier.t;
         position : Position.t;
-        to_next_point : bool
       } [@@deriving yojson]
 
     end
--- a/language-server/vscoqtop/lspManager.ml
+++ b/language-server/vscoqtop/lspManager.ml
@@ -41,6 +41,8 @@ let max_memory_usage  = ref 4000000000
 let full_diagnostics = ref false
 let full_messages = ref false
 
+let point_interp_mode = ref Settings.PointInterpretationMode.Cursor
+
 let block_on_first_error = ref true
 
 let Dm.Types.Log log = Dm.Log.mk_log "lspManager"
@@ -130,7 +132,8 @@ let do_configuration settings =
   full_diagnostics := settings.diagnostics.full;
   full_messages := settings.goals.messages.full;
   max_memory_usage := settings.memory.limit * 1000000000;
-  block_on_first_error := settings.proof.block
+  block_on_first_error := settings.proof.block;
+  point_interp_mode := settings.proof.pointInterpretationMode
 
 let send_configuration_request () =
   let id = `Int conf_request_id in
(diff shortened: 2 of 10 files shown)


Durbatuluk1701:
Another possibility I have considered is that the `getNextPosition` function could be utilized in continuous mode to help prevent repeatedly sending `interpretToPoint` requests that won't change the proof state since they are mid tactic.

For example (where `|` represents a cursor):
`ta|c1. tac2` transitioning to `tac|1. tac2`
Should not need a new `interpretToPoint` request as no proof state change would occur.

rtetley:
I like this ! However I wonder if finding the correct position should be done on the server side rather. What do you think @gares ?

rtetley:
Also I think it should be an option maybe other users can give their sentiment about this ?

gares (changes requested):
I love it too, thanks!

The lang server knows the end of sentences, it should compute what next means.

SkySkimmer:
>This PR adds a "interpret to next point" that will take the current cursor position and if it is part of a tactic then it will interpret through the completion of that tactic.

Is it actually tactic specific? I guess it probably works for any command (eg `Definition foo1 |:= 1. Definition foo2 := 2.`)

Durbatuluk1701:
> > This PR adds a "interpret to next point" that will take the current cursor position and if it is part of a tactic then it will interpret through the completion of that tactic.
> 
> Is it actually tactic specific? I guess it probably works for any command (eg `Definition foo1 |:= 1. Definition foo2 := 2.`)

You are correct it isn't tactic specific. I think my wording is definitely wrong, but I think any solution that would fix #870 would probably need to be general/work for any command and not tactic specific.

Durbatuluk1701:
> I love it too, thanks!
> 
> The lang server knows the end of sentences, it should compute what next means.

Sounds good, I can start working on that. 

One point of clarification, do we want a separate "interpretToNextPoint" sort of command (that would be utilized by continuous mode), or to modify the existing "interpretToPoint" functionality to interpret to the next command?

I think the primary difference would be how using "interpretToPoint" works when in manual mode.

rtetley:
> > I love it too, thanks!
> > The lang server knows the end of sentences, it should compute what next means.
> 
> Sounds good, I can start working on that.
> 
> One point of clarification, do we want a separate "interpretToNextPoint" sort of command (that would be utilized by continuous mode), or to modify the existing "interpretToPoint" functionality to interpret to the next command?
> 
> I think the primary difference would be how using "interpretToPoint" works when in manual mode.

Yes I think we should create a ```interpretToNextPoint``` function in the server. We then create a boolean server setting that can be toggled from the front end. Depending on that setting a ```vscoq/interpretToPoint``` request calls one or the other.

Durbatuluk1701:
Pushed some changes to make it so that the computation occurs on the server side and an option for configuring it.

I was a little confused on all the times that `id_of_pos` was being called and when some of those should have the `next_pos` switch hard-coded to false vs. when it should be propagated from above, so a keen eye reviewing some of those would be helpful.

I additionally encountered a corner case that I don't think has been addressed in the previous discussion in regard to a case with bullets like:
```
Lemma test : forall {A} (x y : A), x = x /\ y = y.
Proof.
  intros.
  split.
  - eauto.
  - eauto.
Qed.
```

Specifically for bullets, do we want `-| eauto.` to interpret to:
1. `-| eauto.` (bullets break up commands)
2. `split.| \n - eauto.` (bullets don't break up commands, roll back before any bullets)
3. `- eauto.|` (bullets don't break up commands, roll forward after the bullet)

I currently have it set up to do option 1.

TheoWinterhalter:
Option 1 is how I use interpret to point most of the time so it would be my favoured option as well.

language-server/vscoqtop/lspManager.ml
```diff
@@ -43,6 +43,8 @@ let full_messages = ref false
 
 let block_on_first_error = ref true
 
+let interp_to_next_position = ref true
```
rtetley, line 46: Instead of making this an option in the backend, I would rather keep the option in the frontend and make a new lsp request called ```vscoq/interpretToNextPoint```. This would need to be implemented in the extProtocol.ml file.
Then depending on the option in the frontend, if the user launches a ```Coq: Interpret To Point``` command it either launches a ```vscoq/interpretToPoint``` or ```vscoq/interpretToNextPoint```:
```typescript
export const sendInterpretToPoint = (editor: TextEditor,  client: Client) => {
    const textDocument = makeVersionedDocumentId(editor);
    const position = editor.selection.active;
    const mode = workspace.getConfiguration("vscoq.proof").pointInterpretationMode;
    if(mode === 0) {
        client.sendNotification("vscoq/interpretToPoint", {textDocument: textDocument, position: position});
    } else {
        client.sendNotification("vscoq/interpretToNextPoint", {textDocument: textDocument, position: position});
    }
 };
```

```

File: language-server/vscoqtop/lspManager.ml [hunk 2]
```diff
@@ -384,7 +390,7 @@ let coqtopInterpretToPoint params =
   match Hashtbl.find_opt states (DocumentUri.to_path uri) with
   | None -> log @@ "[interpretToPoint] ignoring event on non existant document"; []
   | Some { st; visible } ->
-    let (st, events, error_range) = Dm.DocumentManager.interpret_to_position st position ~should_block_on_error:!block_on_first_error in
+    let (st, events, error_range) = Dm.DocumentManager.interpret_to_position st position ~should_block_on_error:!block_on_first_error ~next_pos:!interp_to_next_position in
```
rtetley, line 393: From the previous comment, then instead of having this:
- do not modify ```Dm.DocumentManager.interpret_to_position```
- instead create a new function ```coqtopInterpretToNextSentence``` which does something among the lines of:
```ocaml
let coqtopInterpretToNextPoint params =
  let Notification.Client.InterpretToPointParams.{ textDocument; position } = params in
  let position = Dm.DocumentManager.get_position_of_next_sentence position in
  let uri = textDocument.uri in
  (*do all the same stuff as the other function*)
```
- implement the function ```get_position_of_next_sentence``` in the document manager

rtetley (commented):
I think this approach will make the code more readable and avoid possible regressions due to too many API changes and unnecessary boolean flags.

Durbatuluk1701:
> I think this approach will make the code more readable and avoid possible regressions due to too many API changes and unnecessary boolean flags.

Much cleaner approach indeed! Thank you for your help

rtetley:
Yup ! Thanks a lot ! This is good for me ! @gares I'll let you take care of it as you requested the changes !

gares:
I think we can improve further and kill that regexp.

Instead of computing the position of the next sentence on the RawDocument, we can do so on the (parsed) Document.
Here for example we do something similar, see the call to find_sentence_after.

https://github.com/coq-community/vscoq/blob/f8b2c9fbf4c20cd9c156d8134333e943621c742f/language-server/dm/documentManager.ml#L326

I this the easiest would be keep just one `coqtopInterpretToPoint` with a boolean flag and then call `Dm.DocumentManager.interpret_to_position` or a new `Dm.DocumentManager.interpret_to_next_position`. The two functions would be almost identical (but they are 2 lines each) and the latter would call a new `id_of_sentence_after_pos` that should look like `get_next_range` (the code I link above).

WDYT @rtetley , am I missing something? I don't have the time to try this solution out, but looks simpler.

rtetley:
The problem is that all the other functions in the interpretToPoint function rely on the position, such as ```mk_proof_view_event```, etc.... So you at least have to propagate the new position. You solution would be fine as long as ```Dm.DocumentManager.interpret_to_next_position``` also returns the value of ```next_pos```, and the we use that for the rest of the functions.

gares:
Look better to me. @rtetley would it be possible to make the boolean argument optional in the protocol so to be backwards compatible?

language-server/protocol/extProtocol.ml
```diff
@@ -26,6 +26,7 @@ module Notification = struct
       type t = {
         textDocument : VersionedTextDocumentIdentifier.t;
         position : Position.t;
+        to_next_point : bool
       } [@@deriving yojson]
```
rtetley, line 30: > Look better to me. @rtetley would it be possible to make the boolean argument optional in the protocol so to be backwards compatible?

I believe it is enough to make ```to_next_point``` an option. i.e.:
```suggestion
      type t = {
        textDocument : VersionedTextDocumentIdentifier.t;
        position : Position.t;
        to_next_point : bool option;
      } [@@deriving yojson]
```

rtetley: But I guess there is a larger question here, and I feel kinda silly because I am the one that suggested making this part of the protocol instead of the server options. Are we going to use both versions of simultaneously ? If it makes sense to sometimes use the ```to_next_point: true``` version and false in the same document we should keep it that way. Otherwise we can revert to making it a server option. What do you think @gares ?

gares: I see no use case of using both behaviors at the same time :-/

rtetley: Well I think then we should revert to making it an option in the server that gets sent through the configuration request ! Sorry @Durbatuluk1701 😅

Durbatuluk1701: All good, I think the iterations has made this an overall better solution for sure

language-server/vscoqtop/lspManager.ml
```diff
@@ -382,9 +385,16 @@ let coqtopInterpretToPoint params =
...
   | Some { st; visible } ->
-    let (st, events, error_range) = Dm.DocumentManager.interpret_to_position st position ~should_block_on_error:!block_on_first_error in
+    let (st, events, error_range, position) =
+      match !point_interp_mode with
+      | Settings.PointInterpretationMode.Cursor ->
+        let st, evs, blocking_err = Dm.DocumentManager.interpret_to_position st position ~should_block_on_error:!block_on_first_error in
+        (st, evs, blocking_err, position)
+      | Settings.PointInterpretationMode.NextCommand ->
+        Dm.DocumentManager.interpret_to_next_position st position ~should_block_on_error:!block_on_first_error
+    in
```
Durbatuluk1701, line 397: Do we want to be doing the logic on the `point_interp_mod` ref here, or should we basically reduce it to a bool in `do_configuration` instead?

rtetley: The way its done here looks perfect to me !

rtetley (approved):

rtetley:
@gares I can't merge this since you asked for the changes !

gares (approved):

PR #899, opened 2024-09-11
Fix Hover for period qualifiers and single quotes in names

### LINKED ISSUE #884: Hover on qualifed names
In a window with the following contents:

```
Module sandbox.

Lemma Earth_shadowing_result : 2 + 2 = 4.
Proof. reflexivity. Qed.

End sandbox.

Lemma unqualified_name : 2 + 2 = 4.
Proof. exact sandbox.Earth_shadowing_result. Qed.

Lemma simple_corollary : 2 <= 4.
Proof.
rewrite <- unqualified_name.

```
When hovering with the mouse on the theorem name "sandbox.Earth_shadowing_result" (8th line) no extra information is displayed.  When hovering with the mouse on the name unqualified_name, the type of this theorem appears.  The first hovering result should be the same as the second one.

Durbatuluk1701:
This PR modified the `word_at_position` function to allow for `.` and `'` to appear in names.

This should fix #884 

One note: It does allow for "words" to begin with a `.`, however it is just handled by providing no hover information since it would not be a legal name in Coq.

### DIFF
--- a/language-server/dm/rawDocument.ml
+++ b/language-server/dm/rawDocument.ml
@@ -87,7 +87,7 @@ let range_of_loc raw loc =
   }
 
 let word_at_position raw pos : string option =
-  let r = Str.regexp {|\([a-zA-Z_][a-zA-Z_0-9]*\)|} in
+  let r = Str.regexp {|\([a-zA-Z_.][a-zA-Z_0-9.']*\)|} in
   let start = ref (loc_of_position raw pos) in
   let word = ref None in
   while (start.contents >= 0 && Str.string_match r raw.text start.contents) do


rtetley:
@gares Do you think this could possibly create unwanted side effects ?

gares:
I think that adding `'` is a net win, I've seen quite some code naming stuff like `foo'`.

The `.` is a bit unclear to me. `Nat.pl|us` is today considered as `plus`, that is probably bad since it is more ambiguous than `Nat.plus`. At the same time `Na|t.plus` counts a `Nat`. I don't know if, in that case, if the user expects to get info on the module `Nat` or the addition. (I use `|` for the caret position)

rtetley:
I think until we introduce some form of namespace it is a matter of deciding which is worse:
- Not getting hover information on ```Nat.pl|us```
- Getting ```Nat.plus``` info on ```Na|t.plus```

language-server/dm/rawDocument.ml
```diff
@@ -87,7 +87,7 @@ let range_of_loc raw loc =
   }
 
 let word_at_position raw pos : string option =
-  let r = Str.regexp {|\([a-zA-Z_][a-zA-Z_0-9]*\)|} in
+  let r = Str.regexp {|\([a-zA-Z_.][a-zA-Z_0-9.']*\)|} in
```
gares, line 90: I don't get why the . need to be part of the first character set.
The code below tries the regexp at a position and then goes back until it can.

Durbatuluk1701: I originally didn't have it there but encountered the following issue since the code rolls back a single character at a time:
A case like `Nat.pl|us` it would roll back to `Nat|.plus` which then would no longer match unless the . was in first character set.

If you want I think a more robust way would be to roll back as far as possible to until you encounter `[^a-zA-Z_0-9.']` then consume forward with `{|\([a-zA-Z_][a-zA-Z_0-9.']*\)|}`

gares: then there is an off by one error in the previous code

gares:
I think we can merge, but I did not test it.

language-server/dm/rawDocument.ml
Durbatuluk1701: Hmm yes I do think some more errors could creep up. If you hover something like `lemma1_te|st` I have a feeling it would stop at `lemma|1_test` since `[0-9]` does not appear in the first character set

Durbatuluk1701: Just confirmed this is the case, will rework to be more robust

Durbatuluk1701:
Pushed a rework that should handle cases like:
`lemma1_te|st`
`sandbox.mylem|ma` (also `sand|box.mylemma` will have the same results, still not sure if that is desired behavior)
`lem|ma'` (fully matches `lemma'` and does not drop at single quotes)

rtetley:
We did some experimenting with @gares and were able to get a solution to the Module.foo problem, I'll merge as soon as CI goes green ! Thanks again !

language-server/dm/rawDocument.ml
```diff
@@ -87,14 +87,20 @@ let range_of_loc raw loc =
...
-  word.contents
+  try
+    let back_reg = Str.regexp {|[^a-zA-Z_0-9.']|} in
+    let start_ind = loc_of_position raw pos in
+    (* Search backwards until we find a character that cannot be part of a word *)
+    let first_non_word_ind = Str.search_backward back_reg raw.text start_ind in
+    let forward_reg = Str.regexp {|[^a-zA-Z_0-9']|} in
+    (* Search forwards ensuring that all characters are part of a well defined word. (Cannot start with [0-9'.] and cannot end with .)*)
+    let last_word_ind = Str.search_forward forward_reg raw.text start_ind in
+    let word = String.sub raw.text first_non_word_ind (last_word_ind - first_non_word_ind) in
```
gares, line 98: still looks like an off by one @rtetley .
Did we try `(Foo.bar` or the like?

Durbatuluk1701: Nice catch, think #903 will fix

PR #900, opened 2024-09-11
feat: optimize pp-display for a better and faster goal view and query panel

### LINKED ISSUE #752: Bad alignment in the output of "Print Universes Subgraph"
A minor QoL nitpick, but in the following code:

```coq
Universe u v.
Print Universes Subgraph (u v).
```

Vscoq displays:

```
Set < u
< v
```
while proof general displays the easier to parse aligned version:
```
Set < u
    < v
```

### LINKED ISSUE #835: Unsatisfactory line break
In the attached picture the formatting of the last n - 2 is weird.
Screenshot at 2024-07-18 14-12-24

### LINKED ISSUE #859: Give clear feedback when a proof view is busy computing
Following this [topic](https://coq.zulipchat.com/#narrow/stream/237662-VsCoq-devs-.26-users/topic/Light.20green.20highlighting.20gone.3F).

We would like to add a spinner or another UI element to indicate that the goal view is busy computing.

### LINKED ISSUE #869: Significant delay between tactic completion and goal space update
I just upgraded from VSCoq 0.3.9 to 2.1.7, and I've noticed a significant delay (~1 second) between the completion of some tactics and the re-render of the goal space in the goal window. I'm on Coq .19.0, NixOS, coq and the language server installed via opam.

rtetley:
This PR aims at making a better pretty printing algo for the goal view.
It is heavily inspired by the [Oppen algorithm](http://i.stanford.edu/pub/cstr/reports/cs/tr/79/770/CS-TR-79-770.pdf).
Closes #859 
Closes #869 
Closes #752
Closes #835

### DIFF
--- a/client/pp-display/src/pp.tsx
+++ b/client/pp-display/src/pp.tsx
@@ -1,14 +1,12 @@
-import { FunctionComponent, ReactFragment, useRef, useState, useEffect, useLayoutEffect} from 'react';
+import { FunctionComponent, useRef, useState, useEffect, useLayoutEffect} from 'react';
 import useResizeObserver from '@react-hook/resize-observer';
 import {ResizeObserverEntry} from '@juggle/resize-observer';
 import { v4 as uuid } from 'uuid';
 
-import { PpString, PpMode, BoxDisplay, Term, Break, Box, DisplayType, BreakInfo, HideStates } from './types';
+import { PpString, PpMode, BoxDisplay, Term, Break, Box, DisplayType, BreakInfo, HideStates, Token, TokenType } from './types';
 import PpBox from './pp-box';
 
 import classes from './Pp.module.css';
-
-const MAX_RECOMPUTE = 50;
 
 type PpProps = {
     pp: PpString;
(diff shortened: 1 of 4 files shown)


gares:
Funny that half of the citations in that paper are by Gerard Huet and Gilles Kahn...

client/pp-display/src/pp.tsx
```diff
@@ -262,82 +263,129 @@ const ppDisplay : FunctionComponent<PpProps> = (props) => {
...
-                        breakInfo = {id: breakId, offset: offset};
+    const estimateBoxWidth = (box: Box, context: CanvasRenderingContext2D) : number => {
+        let currentWidth = 0;
+        for (let childBox of box.boxChildren) {
+            if(childBox) {
+                switch(childBox.type) {
+                    case DisplayType.box:
+                        currentWidth += estimateBoxWidth(childBox, context);
+                        break;
+                    case DisplayType.break:
```
tomtomjhj, line 292: I haven't tried out this, but I believe the estimation would be more accurate if
* a `Break` with `shouldBreak: true` resets the current width
* `estimateBoxWidth` returns the max width

rtetley: I am not sure what you mean ?
Currently the estimation of whether to break or not does not happen here but rather in the ```scanTokenStream``` function.
The box always has the max width but its only relevant to breaks if a box is a ```hvBox``` in which case all its breaks should happen.
Otherwise we use the estimation of a ```block``` which happens from break to break.

rtetley: Yeah, I agree there is definitely a more optimal way to do this. However ```buildTokenStream``` is only called once on receiving a ```ppString``` and in practice, even on examples that use to take a considerable amount of time, this version speeds everything up by an order of magnitude. I think there will be time later down the road to optimise this further.

PR #901, opened 2024-09-12
feat: show message when goals are unfocused

### LINKED ISSUE #643: Feature request: showing the next bullet
Hi. I'm using VsCoq, and it's pretty nice! However, I think it'd be great if VsCoq prints the next bullet on the completion of proof for a bullet.

Currently, VsCoq only says "No more subgoals".
29-09-2023_19:09:46_screenshot

[Coqtail](https://github.com/whonore/Coqtail), a Vim plugin for Coq, has the feature like this.
29-09-2023_19:08:19_screenshot

I'm using VsCoq 2.0.1 and vscoq-language-server 2.0.1+coq8.18.

rtetley:
When there are unfocused goals (when bullets are used) we now display a message stating the subproof is complete but there are still unfocused goals.
Closes #643

### DIFF
--- a/client/goal-view-ui/src/components/templates/ProovViewPage.tsx
+++ b/client/goal-view-ui/src/components/templates/ProovViewPage.tsx
@@ -50,12 +50,13 @@ const proofViewPage: FunctionComponent<ProofViewPageProps> = (props) => {
                 <GoalSection 
                     key={"goals"}
                     goals={goals!.main}
-                    collapseGoalHandler={(id) => collapseGoalHandler(id, ProofViewGoalsKey.main)} 
+                    unfocusedGoals={goals!.unfocused}
+                    collapseGoalHandler={(id) => collapseGoalHandler(id, goals!.main.length ? ProofViewGoalsKey.main : ProofViewGoalsKey.unfocused)}
                     displaySetting={displaySetting}
                     emptyMessage={
                         goals!.shelved.length ? "There are shelved goals. Try using `Unshelve.`" :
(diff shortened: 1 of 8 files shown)


client/goal-view-ui/src/components/templates/ProovViewPage.tsx
```diff
@@ -49,12 +49,13 @@ const proofViewPage: FunctionComponent<ProofViewPageProps> = (props) => {
...
                 <GoalSection 
                     key={"goals"}
-                    goals={goals!.main} 
+                    goals={goals!.main}
                     collapseGoalHandler={(id) => collapseGoalHandler(id, ProofViewGoalsKey.main)} 
                     displaySetting={displaySetting}
                     emptyMessage={
                         goals!.shelved.length ? "There are shelved goals. Try using `Unshelve.`" :
                         goals!.givenUp.length ? "There are some goals you gave up. Go back and solve them, or use `Admitted.`" :
+                        goals!.unfocused.length ? "The subproof is complete, but there are some unfocused goals. Focus next goal with bullet -." :
```
Durbatuluk1701, line 58: Is it maybe a bit confusing to specify `-` as the bullet type to use when it could be any of them

gares: Yes, I think we had a discussion with @SkySkimmer about the API to use here (or to suggest the right bullet as a quickfix if the wrong one is used)

rtetley: I'll change this to make the bullet suggestion non specific and I'll display the unfocused goals as well

Issue #731, opened 2024-02-15
Valid but non-ascii identifiers are not processed?


YaZko:
A fairly niche shortcoming, but Coq allows for a little bit more than ASCII in identifiers, which VSCoq doesn't seem to take into account. For instance in the following minimal file, hovering over `Y` gives information about it, but not in the case of `Xᵒ`.

```coq
Definition Xᵒ : nat := 0.

Definition Y : nat := 1.

(* In the following, vscoq gives feedback about Y but not Xᵒ *)
Goal Y + Xᵒ = 1.
```

PR #903, opened 2024-09-12
fix: off by one error in Hover

Durbatuluk1701:
Changes the string we are hovered on to be the start of the word index to end of word (does not include any period at the end)

### DIFF
--- a/language-server/dm/rawDocument.ml
+++ b/language-server/dm/rawDocument.ml
@@ -99,7 +99,7 @@ let word_at_position raw pos : string option =
     (* we get the substring from the first word index to the last index for the word *)
     let word = String.sub raw.text first_word_ind (last_word_ind - first_word_ind) in
     Some word
-  with Not_found ->
+  with _ ->
     None
 
 let apply_text_edit raw (Range.{start; end_}, editText) =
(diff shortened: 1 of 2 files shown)


language-server/dm/rawDocument.ml
```diff
@@ -92,15 +92,15 @@ let word_at_position raw pos : string option =
...
+    let first_word_ind = first_non_word_ind + 1 in
     let forward_reg = Str.regexp {|[^a-zA-Z_0-9']|} in
     (* Search forwards ensuring that all characters are part of a well defined word. (Cannot start with [0-9'.] and cannot end with .)*)
     let last_word_ind = Str.search_forward forward_reg raw.text start_ind in
-    let word = String.sub raw.text first_non_word_ind (last_word_ind - first_non_word_ind) in
+    (* we get the substring from the first word index to the last index for the word *)
+    let word = String.sub raw.text first_word_ind (last_word_ind - first_word_ind) in
     Some word
-  with Not_found ->
+  with _ ->
```
Durbatuluk1701, line 102: Not sure if this is idiomatic in Ocaml, but I noticed when looking at #731 that if we attempt to hover on a non-ascii character it could cause a crash due to `String.sub` raising `Invalid_argument`

Durbatuluk1701: More generally I wonder (I am not really experienced in Ocaml) what the proper way for doing something akin to Regex in UTF would be to help eventually solve 731.

rtetley: Not sure how to handle this. Maybe @gares has pointers ?

rtetley:
Thanks !

PR #904, opened 2024-09-13
feat: add config option to control whether Coq buttons are displayed

Durbatuluk1701:
I personally don't like many buttons in the Editor Actions tab menu, so I think it would be nice to be able to disable them from appearing.
When `vscoq.general.display-buttons == true` (which it is by default) the buttons will be visible:
image
Otherwise, the buttons won't appear.

I could not really think of a good pre-existing setting scope to put them in, so I created one called `general`, but if anyone has renaming thoughts or a better place to put it just feel free.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -106,18 +106,6 @@
         }
       },
       {
-        "title": "General",
-        "type": "object",
-        "properties": {
-          "vscoq.general.display-buttons": {
-            "scope": "workspace",
-            "type": "boolean",
-            "default": true,
-            "description": "Toggles whether the Coq buttons are displayed in the editor actions tab menu."
-          }
-        }
-      },
-      {
         "title": "Memory management",
         "type": "object",
         "properties": {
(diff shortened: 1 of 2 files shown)


client/package.json
```diff
@@ -105,6 +105,18 @@
...
+        "type": "object",
+        "properties": {
+          "vscoq.general.display-buttons": {
+            "scope": "workspace",
+            "type": "boolean",
+            "default": true,
+            "description": "Toggles whether the Coq buttons are displayed in the editor actions tab menu."
+          }
+        }
+      },
```
rtetley, line 119: I would make this part of the "Proof checking scope" as ```vscoq.proof.display-buttons```

rtetley (changes requested):

Durbatuluk1701:
@rtetley Can I get some guidance on why this CI is failing?

rtetley:
Just looks like a sporadic network error failure.

rtetley (approved):
Thanks !

PR #911, opened 2024-09-20
feat: jump to definition proof of concept

rtetley:
WIP

cf https://github.com/coq/coq/pull/19584

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,7 +4,7 @@
   inputs = {
     flake-utils.url = "github:numtide/flake-utils";
 
-    coq-master = { url = "github:rtetley/coq/d6dd8a618cbf628a6580c9a42db0cf5f44606bea"; }; # Should be kept in sync with PIN_COQ in CI workflow
+    coq-master = { url = "github:coq/coq/d6dd8a618cbf628a6580c9a42db0cf5f44606bea"; }; # Should be kept in sync with PIN_COQ in CI workflow
     coq-master.inputs.nixpkgs.follows = "nixpkgs";
 
   };
--- a/flake.nix
+++ b/flake.nix
@@ -4,7 +4,7 @@
   inputs = {
     flake-utils.url = "github:numtide/flake-utils";
 
-    coq-master = { url = "github:coq/coq/668dc3ca9735a2b008f5ce52c14969af59cc24e1"; }; # Should be kept in sync with PIN_COQ in CI workflow
+    coq-master = { url = "github:rtetley/coq/d6dd8a618cbf628a6580c9a42db0cf5f44606bea"; }; # Should be kept in sync with PIN_COQ in CI workflow
     coq-master.inputs.nixpkgs.follows = "nixpkgs";
 
   };
--- a/language-server/dm/documentManager.ml
+++ b/language-server/dm/documentManager.ml
@@ -761,42 +761,33 @@ let jump_to_definition st pos =
   | None -> log "jumpToDef: no word found at cursor"; None
   | Some pattern ->
     log ("jumpToDef: found word at cursor: \"" ^ pattern ^ "\"");
-    match st.observe_id with
-    | Top -> log "jumpToDef with no context"; None
-    | (Id id) ->
-      match Document.get_sentence st.document id with
-      | None -> log "jumpToDef: observe_id does not exist"; None
-      | Some { stop } ->
-        let o_pos = RawDocument.position_of_loc raw_doc stop in
-        match get_context st o_pos with
-        | None -> log "No context found"; None
-        | Some _ ->
-          match parse_entry st loc (Procq.Prim.smart_global) pattern with
-          | { v = AN qid } ->
-            let ref = Nametab.locate qid in
-             begin match Nametab.cci_src_loc (TrueGlobal ref) with
-                | None -> None
-                | Some loc ->
-                  begin match loc.Loc.fname with
-                    | Loc.ToplevelInput | InFile  { dirpath = None } -> None
-                    | InFile { dirpath = Some dp } ->
-                        let f = Loadpath.locate_absolute_library @@ Libnames.dirpath_of_string dp in
-                        begin match f with
-                          | Ok f ->
-                            let f =  Filename.remove_extension f ^ ".v" in
-                            (if Sys.file_exists f then
-                              let b_pos = Position.create ~character:(loc.bp - loc.bol_pos) ~line:(loc.line_nb - 1) in
-                              let e_pos = Position.create ~character:(loc.ep - loc.bol_pos) ~line:(loc.line_nb - 1) in
-                              let range = Range.create ~end_:b_pos ~start:e_pos in
-                              Some (range, f)
-                            else
-                              None
-                            )
-                          | Error _ -> None
-                        end
+    try
+    let qid = parse_entry st loc (Procq.Prim.qualid) pattern in
+      let ref = Nametab.locate_extended qid in
+        match Nametab.cci_src_loc ref with
+          | None -> None
+          | Some loc ->
+            begin match loc.Loc.fname with
+              | Loc.ToplevelInput | InFile  { dirpath = None } -> None
+              | InFile { dirpath = Some dp } ->
+                  let f = Loadpath.locate_absolute_library @@ Libnames.dirpath_of_string dp in
+                  begin match f with
+                    | Ok f ->
+                      let f =  Filename.remove_extension f ^ ".v" in
+                      (if Sys.file_exists f then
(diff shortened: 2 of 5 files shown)


gares:
Changes to coq look good too.
I wonder if there is a way to circumvent jumping to dune's copy of the sources.
I mean, should the logic be in Coq or here?
@SkySkimmer added the print of the Loc to a user visible message, so I guess it should be in Coq

SkySkimmer:
> @SkySkimmer added the print of the Loc to a user visible message, so I guess it should be in Coq

That logic is on the loc -> pp side though, vscoq only calls the qualid -> loc part from coq and does the rest on its own.

language-server/dm/documentManager.ml
```diff
@@ -658,6 +658,69 @@ let hover st pos =
...
+    | Some Top -> log "jumpToDef with no context"; None
+    | Some (Id id) ->
+      match Document.get_sentence st.document id with
+      | None -> log "jumpToDef: observe_id does not exist"; None
+      | Some { stop } ->
+        let o_pos = RawDocument.position_of_loc raw_doc stop in
+        match get_context st o_pos with
+        | None -> log "No context found"; None
+        | Some _ ->
+          match parse_entry st loc (Pcoq.Prim.smart_global) pattern with
```
SkySkimmer, line 696:
```suggestion
          match parse_entry st loc Pcoq.Prim.qualid pattern with
```
since you're only using the `AN` case of smart_global

gares:
It may still make sense to have the api on the coq side and let vscoq call it on the loc, no?

Anyway, good job!

TheoWinterhalter:
I guess in the meantime something that works in most cases but not all is better than nothing?

language-server/dm/documentManager.ml
```diff
@@ -658,6 +658,71 @@ let hover st pos =
...
+          | { v = AN qid } -> 
+            begin match Nametab.locate qid with
+              | ConstRef x -> begin match Declare.get_loc x with
```
SkySkimmer, line 699: this will need updating as I moved get_loc and made it apply to more than constants
something like
```suggestion
            let extref = Nametab.locate_extended qid in
            match Nametab.cci_src_loc extref with
```
should work

SkySkimmer:
Not sure how well this works with an editable document, I mean start with
~~~coq
(* some comment *)
Definition foo := 0.

Check foo.
~~~
evaluate the Definition, then edit the comment, then jump to def of the `foo` in Check

if editing the comment doesn't force re-evaluating the Definition then the loc will be wrong, but we don't want to re execute after editing comments (or whatever other edits we can detect as being trivial)

gares:
That is certainly true, although it seems a minor problem in practice.

I'm thinking loud here: if we manage to perform some pieces of the globalization phase at parse time, then parsing "Definition foo := ..." could tell us that we will generate `Some.Logical.Path.foo` and we could store the KN->loc map in a synterp summary (rather than an interp one). We could complement the map at interp time with interp-time-generated stuff we really want to be precise, and that part would need to be shifted/shiftable by the UI when it tries to reuse an old state on a new text. Exactly as we do for diagnostics.

In a way, since the map is about locs, it should be part of the parsing state, not the interp one.

SkySkimmer:
>I'm thinking loud here: if we manage to perform some pieces of the globalization phase at parse time

That's impossible IMO

>In a way, since the map is about locs, it should be part of the parsing state, not the interp one.

The map is about objects like globrefs which are part of the interp state.

gares:
> That's impossible IMO

AFAIK synterp already computes the current modpath. From that one can compute the final KN of "foo". Am I missing something?

rtetley:
@gares @SkySkimmer could you review before merge ?

flake.nix
```diff
@@ -4,7 +4,7 @@
   inputs = {
     flake-utils.url = "github:numtide/flake-utils";
 
-    coq-master = { url = "github:coq/coq/668dc3ca9735a2b008f5ce52c14969af59cc24e1"; }; # Should be kept in sync with PIN_COQ in CI workflow
+    coq-master = { url = "github:rtetley/coq/d6dd8a618cbf628a6580c9a42db0cf5f44606bea"; }; # Should be kept in sync with PIN_COQ in CI workflow
```
gares, line 7: should not be necessary anymore

language-server/dm/documentManager.ml
```diff
@@ -739,6 +739,57 @@ let hover st pos =
...
+    match st.observe_id with
+    | Top -> log "jumpToDef with no context"; None
+    | (Id id) ->
+      match Document.get_sentence st.document id with
+      | None -> log "jumpToDef: observe_id does not exist"; None
+      | Some { stop } ->
+        let o_pos = RawDocument.position_of_loc raw_doc stop in
+        match get_context st o_pos with
+        | None -> log "No context found"; None
+        | Some _ ->
```
gares, line 763: this is fishy, why we check for a context and then don't use it?
was this preliminary code to use the proof context when available?
At least it deserves a comment

gares (approved):

language-server/dm/documentManager.ml
rtetley: Yeah, I think its just a way to check that we are indeed in an executed context.
If I just jump to def with no execution state than I just get an uncaught Not_found exception. But I suppose I should probably catch the exception and return None in that case. Or check that the sentence at the cursor is executed ?

rtetley:
I still get errors if I try to use Jump to Def in certain cases. For example:
If I try to ```jump_to_def``` on ```True``` I will get the usual: 
```
Error while opening document. Cannot load a library with the same name as the current one
(Stdlib.Init.Datatypes).
```
 Was that not fixed ?

language-server/dm/documentManager.ml
```diff
@@ -739,6 +739,48 @@ let hover st pos =
...
+  let raw_doc = Document.raw_document st.document in
+  let loc = RawDocument.loc_of_position raw_doc pos in
+  let opattern = RawDocument.word_at_position raw_doc pos in
+  match opattern with
+  | None -> log "jumpToDef: no word found at cursor"; None
+  | Some pattern ->
+    log ("jumpToDef: found word at cursor: \"" ^ pattern ^ "\"");
+    try
+    let qid = parse_entry st loc (Procq.Prim.qualid) pattern in
+      let ref = Nametab.locate_extended qid in
```
SkySkimmer, line 756: I guess this and maybe the locate_absolute_library are where Not_found can come from

rtetley: Yeah I just enclosed it in a big ```try [...] with``` and it no longer breaks the server so that's okay

SkySkimmer:
> If I try to jump_to_def on True I will get the usual:

in the vscoq log do you see `No project file found for .../Logic.v` or `Arguments from project file ...`?

rtetley:
> > If I try to jump_to_def on True I will get the usual:
> 
> in the vscoq log do you see `No project file found for .../Logic.v` or `Arguments from project file ...`?

Yes you are right. I get ```[lspManager          , 49641, 1733315036.425675] No project file found for /Users/rtetley/.opam/vscoq-dev/lib/coq/theories/Init/Datatypes.v
```

SkySkimmer:
Right, we don't install _CoqProject so there's nothing to find.
https://github.com/HoTT/Coq-HoTT/issues/2146 is a similar issue
also https://github.com/coq/coq/issues/15924

rtetley:
> Right, we don't install _CoqProject so there's nothing to find. [HoTT/Coq-HoTT#2146](https://github.com/HoTT/Coq-HoTT/issues/2146) is a similar issue also [coq/coq#15924](https://github.com/coq/coq/issues/15924)

I'm not sure I understand how that pertains to the error I find ? I was under the impression that this is the error you get if you load it twice ? Unless you mean the ```_CoqProject``` contains a -no-init ?

rtetley:
> > Right, we don't install _CoqProject so there's nothing to find. [HoTT/Coq-HoTT#2146](https://github.com/HoTT/Coq-HoTT/issues/2146) is a similar issue also [coq/coq#15924](https://github.com/coq/coq/issues/15924)
> 
> I'm not sure I understand how that pertains to the error I find ? I was under the impression that this is the error you get if you load it twice ? Unless you mean the `_CoqProject` contains a -no-init ?

Answered my own question... https://github.com/coq/coq/blob/master/theories/Init/_CoqProject

gares:
Yes, but we should be sure someone is fixing it ..

rtetley:
> Yes, but we should be sure someone is fixing it ..

Agreed, just double checking !

rtetley:
> Yes, but we should be sure someone is fixing it ..

Should I just merge in the mean time ?

gares:
I guess so

rtetley:
I'll merge and wait from news in https://github.com/coq/coq/issues/15924

PR #915, opened 2024-09-25
support Restart

### LINKED ISSUE #770: VsCoq v2 unsupported command for Restart and Undo
I just updated my VsCoq to v2.1.2 from v0.3.9 (sorry for the delay) but I found a problem that using Restart and Undo will cause an error message 'Unsupported Command'. It seems that the backtrack tactics are yet to be implemented in v2.1.2. Is that the case?

gares:
Fix #770


rtetley:
@gares what is the status on this ? Will you have time to work on it before the next planned release (around the first week of February) ?

PR #916, opened 2024-09-26
cleanup(EM): delegated tasks are executable sentences

gares:
If an opaque proof contained as Skip (i.e. a syntax error) it was not delegated anyway by the scehduler.

I'm rebasing #915 on top of this one since it makes is much cleaner.

@rtetley please have a look at the code for `id_of_first/last_task`. I think my new code is equivalent, but I don't get in which cases the default would be different than the computed one.


rtetley:
If I remember correctly, the opener and closer referred to the first proof step. This often lead us to not treat the ```Proof.``` or ```Qed.``` lines when coloring in delegation mode ?

PR #923, opened 2024-10-03
Add Prettify and Lint scripts for each sub-project and top level client

Durbatuluk1701:
Main Highlights:
1. Yarn scripts for running the prettifier (in-place prettifies files) and linter
2. Linter made stricter
3. Typechecker made stricter (no explicit typecasting with `as`, tries to force exhaustive switch statements, etc.)
 

I think this should hopefully allow for some development quality of life improvements.

One thing I will mention, it feels duplicative (albeit the most expressive) at the moment to have each sub-project carrying its own prettier, linter, and TS config. 
Should I explore trying to collapse these into one top level config that applies to the sub-projects as well, or let each sub-project configure itself in case it needs more fine-grained control?

Another question: Should enforcing meeting the style standard (`prettier --check .`) be a test/requirement checked in CI build?


Durbatuluk1701:
If the new rules and changes seem good, I think I will run `yarn run prettify:all` and `yarn run lint:all`, fix any lint issues, then hopefully the whole `client` dir will be consistent

shilangyu:
Closing since it has been superseded by #1249 and #1241.

I did not notice when writing those PRs that there has been prior work here, sorry for that. If you are not happy with the setup I have done or have suggestions about how to improve it, please open a new PR. I will be very happy to review it.

PR #926, opened 2024-10-10
doc: update doc after organization change

rtetley:


gares:
I think you could be more drastic in the first paragraph, just state that the minimum version, eg

VsCoq is an extension for [Visual Studio Code](https://code.visualstudio.com/) (VS Code) and [VSCodium](https://vscodium.com/) which provides support for the [Coq Proof Assistant](https://coq.inria.fr/). VSCoq requires Coq version 8.18 or above (if you are running an older version, you should use [vscoq-legacy](https://github.com/coq-community/vscoq-legacy)).

WDYT?

gares:
Or better mention the legacy repo in the second section, where the lowest bound is declared

rtetley:
Yeah sounds good

PR #935, opened 2024-10-28
syntax: Highlight Goal in any context

Lysxia:
Reported by Thomas Lamiaux on Zulip. `Goal` is followed directly by the proposition to prove, which can be any term so, unlike `Theorem` etc., it's practically impossible to use lookarounds to disambiguate the `Goal` command from an identifier.

```
Goal prop.
(* different from *)
Theorem ident : prop.
```


gares (approved):

rtetley:
Just took the liberty of rebasing and will merge after CI

Issue #914, opened 2024-09-25
Slowdown after editing file


TheoWinterhalter:
Since `2.2.0` and also with `2.2.1`, I experience slowdowns whenever I edit the file. It seems the interface takes a few seconds to become responsive again.

Perhaps it's some parsing issue? I currently don't have the time to properly test when it happens, but I get the felling it happens all the time after editing a file that was partially processed.

Issue #682, opened 2023-11-01
Changes appear to be processed character by character (VsCoq2 )


nathanwhit:
Often if I modify a proof, it takes a long time to update the goals view with the changes, but more notably, it appears that the document (at least, the proof being modified and all following proofs) is re-executed for each typed character. In other words, if I type 8 characters, the document is re-executed 8 times. The following video illustrates this fairly well:
https://github.com/coq-community/vscoq/assets/17734409/fc5f7897-9ed4-4a90-af95-faf047926745

The problem is most noticeable in larger files, but I've also encountered it in smaller files.

I would expect that receiving a new `documentDidChange` notification for a given document would cancel any ongoing work handling earlier change notifications for that document, but instead, it almost looks like the change notifications just pile up and get processed sequentially.

Blaisorblade:
I agree with this issue — and even with manual mode, my impression is that the file is re-parsed after every keypress.

And this issue makes VsCoq2 unusable for me at the moment.

PR #937, opened 2024-10-31
Interupt consecutive parsing events

rtetley:
Currently, as outlined in #682, each time a document is edited, a full parsing is triggered. This cam be problematic and could be at the origin of #914.
The aim of this PR is to make parsing part of the event loop which can then be optimised to only keep the last event.


rtetley:
Closed in favour of #941

PR #941, opened 2024-11-06
Cancel parsing event

### LINKED ISSUE #682: Changes appear to be processed character by character (VsCoq2 )

### LINKED ISSUE #914: Slowdown after editing file

rtetley:
Trying another route from #937.
This time we use the SEL api to cancel events.
Currently does not seem to be working because of problem noted in https://github.com/gares/sel/pull/10

Closes #682.
Closes #914.

### DIFF
--- a/language-server/dm/document.mli
+++ b/language-server/dm/document.mli
@@ -37,6 +37,10 @@ type outline_element = {
 
 type outline = outline_element list
 
+type event
+
+type events = event Sel.Event.t list
+
 val raw_document : document -> RawDocument.t
 
 val outline : document -> outline
@@ -45,10 +49,15 @@ val create_document : Vernacstate.Synterp.t -> string -> document
 (** [create_document init_synterp_state text] creates a fresh document with content defined by
     [text] under [init_synterp_state]. *)
 
-val validate_document : document -> sentence_id option * sentence_id_set * document
-(** [validate_document doc] parses the document without forcing any execution
-    and returns the id of the bottommost sentence of the prefix which has not changed
-    since the previous validation, as well as the set of invalidated sentences *)
+val validate_document : document -> events
+(** [validate_document doc] triggers the parsing of the document line by line without
+    launching any execution. *)
+
+val handle_event : document -> event -> events * (sentence_id option * sentence_id_set * document) option
+(** [handle_event dpc ev] handles a parsing event for the document. One parsing event parses one line
+    and prepares the next parsing event. Finally once the full parsing is done, the final event returs
+    the id of the bottomost sentence of the prefix which has not changed since the previous validation
+    as well as the set of invalidated sentences. *)
 
 type parsed_ast = {
   ast: Synterp.vernac_control_entry;
@@ -62,6 +71,8 @@ type parsing_error = {
   msg: string Loc.located;
   qf: Quickfix.t list option;
 }
+
+type parse_state
 
 val parse_errors : document -> parsing_error list
 (** [parse_errors doc] returns the list of sentences which failed to parse
--- a/language-server/dm/document.mli
+++ b/language-server/dm/document.mli
@@ -49,11 +49,11 @@ val create_document : Vernacstate.Synterp.t -> string -> document
 (** [create_document init_synterp_state text] creates a fresh document with content defined by
     [text] under [init_synterp_state]. *)
 
-val validate_document : document -> events
+val validate_document : document -> document * events
 (** [validate_document doc] triggers the parsing of the document line by line without
     launching any execution. *)
 
-val handle_event : document -> event -> events * (sentence_id option * sentence_id_set * document) option
+val handle_event : document -> event -> document * events * (sentence_id option * sentence_id_set * document) option
 (** [handle_event dpc ev] handles a parsing event for the document. One parsing event parses one line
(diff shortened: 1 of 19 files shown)


rtetley:
This is a big one, still WIP. Needs:
- [ ] Recomputing the overview when after a new parse event
- [ ] Fixing the mechanism for getting proof views -> should be handled in the doc manager

language-server/dm/document.mli
```diff
@@ -45,10 +50,15 @@ val create_document : Vernacstate.Synterp.t -> string -> document
...
-    since the previous validation, as well as the set of invalidated sentences *)
+val validate_document : document -> document * events
+(** [validate_document doc] triggers the parsing of the document line by line without
+    launching any execution. *)
+
+val handle_event : document -> event -> document * events * (sentence_id option * sentence_id_set * document * document) option
+(** [handle_event dpc ev] handles a parsing event for the document. One parsing event parses one line
+    and prepares the next parsing event. Finally once the full parsing is done, the final event returs
+    the id of the bottomost sentence of the prefix which has not changed since the previous validation
+    as well as the set of invalidated sentences. *)
```
gares, line 61: I think you should build a record for the return type, so that we can name all these documents

language-server/dm/documentManager.mli
```diff
@@ -38,17 +39,24 @@ val pp_event : Format.formatter -> event -> unit
...
+    events: events;
+    update_view: bool;
+    notification: Notification.Server.t option;
+}
+
+val init : Vernacstate.t -> opts:Coqargs.injection_command list -> DocumentUri.t -> text:string -> state * events
 (** [init st opts uri text] initializes the document manager with initial vernac state
-    [st] on which command line opts will be set. *)
+    [st] on which command line opts will be set. [background] indicates if the document should
+    be executed in the background once it is parsed. *)
```
gares, line 52: there is no background here

language-server/dm/documentManager.mli
```diff
@@ -38,17 +39,24 @@ val pp_event : Format.formatter -> event -> unit
...
+val apply_text_edits : state -> text_edit list -> state * events
+(** [apply_text_edits doc edits] updates the text of [doc] with [edits]. 
+    A ParseEvent is triggered, once processed: the new
     document is parsed, outdated executions states are invalidated, and the observe
-    id is updated. *)
-
-val clear_observe_id : state -> state
-(** [clear_observe_id state] updates the state to make the observe_id None *)
+    id is updated. Finally if [background] is set to true, an execution is launched in
+    the background *)
```
gares, line 59: there is no background here too

language-server/dm/documentManager.mli
```diff
@@ -59,41 +67,41 @@ val get_next_range : state -> Position.t -> Range.t option
...
+val interpret_in_background : state -> should_block_on_error:bool -> (state * events)
 (** [interpret_in_background doc] same as [interpret_to_end] but computation 
     is done in background (with lower priority) *)
 
-val reset : state -> state * events
+val reset : state -> state * events 
 (** resets Coq *)
 
-val executed_ranges : state -> exec_overview
+val executed_ranges : state -> Settings.Mode.t -> exec_overview
```
gares, line 97: it is unclear what the mode is for here

gares (approved):
looks like a great cleanup

language-server/dm/document.mli
```diff
@@ -45,10 +58,15 @@ val create_document : Vernacstate.Synterp.t -> string -> document
...
-    since the previous validation, as well as the set of invalidated sentences *)
+val validate_document : document -> document * events
+(** [validate_document doc] triggers the parsing of the document line by line without
+    launching any execution. *)
+
+val handle_event : document -> event -> document * events * parsing_end_info option
+(** [handle_event dpc ev] handles a parsing event for the document. One parsing event parses one line
+    and prepares the next parsing event. Finally once the full parsing is done, the final event returs
+    the id of the bottomost sentence of the prefix which has not changed since the previous validation
+    as well as the set of invalidated sentences. *)
```
gares, line 69: When parsing ends, we still have two documents, hence two cancelation handles. This is a bit confusing: which document should the client keep?

rtetley: Not sure I catch your drift ? We have one "parsing" document which just became "parsed", and one "previous" document which is the version before all the new parsing events happened.

PR #944, opened 2024-11-12
Support per-file _CoqProject

SkySkimmer:
cf https://github.com/coq/coq/pull/19826

### DIFF
--- a/language-server/vscoqtop/lspManager.ml
+++ b/language-server/vscoqtop/lspManager.ml
@@ -23,7 +23,7 @@ open Protocol.ExtProtocol
 
 module CompactedDecl = Context.Compacted.Declaration
 
-let init_state : (Vernacstate.t * Coqargs.injection_command list) option ref = ref None
+let init_state : Vernacstate.t option ref = ref None
 let get_init_state () =
   match !init_state with
   | Some st -> st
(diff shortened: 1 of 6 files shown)


language-server/vscoqtop/lspManager.ml
```diff
@@ -257,9 +257,28 @@ let reset_observe_ids =
...
+      Coqargs.default
+    | Some f ->
+      let project = CoqProject_file.read_project_file ~warning_fn:(fun _ -> ()) f in
+      let args = CoqProject_file.coqtop_args_from_project project in
+      log (Printf.sprintf "Arguments from project file %s: %s" f (String.concat " " args));
+      fst @@ Coqargs.parse_args ~init:Coqargs.default args
+  in
+
+  let () = Vernacstate.unfreeze_full_state vst in
+  let () = Coqinit.init_document local_args in
```
SkySkimmer, line 277: TODO optcomp
I guess for older coq we do nothing as init_runtime includes init_document

(we still get injections from the local _CoqProject so -noinit works but basically everything else is from the root coqproject)

SkySkimmer:
Wrong branch name, use https://github.com/coq/vscoq/pull/945 instead

PR #945, opened 2024-11-15
Support per-file _CoqProject

SkySkimmer:
cf https://github.com/coq/coq/pull/19826


gares:
@rtetley CI was not very green, but at least on mster it should work

PR #957, opened 2024-12-04
Fixing the outline

### LINKED ISSUE #953: Outline regression
It seems no longer works consistently. Probably due to a race condition between the multiple parse events and querying the document to get the outline ? Should investigate.

rtetley:
Closes #953


rtetley:
I have explored several options:
- First I tried to implement it using a partialResult or workDone tokens (see [here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResults)). But vscode does not seem to send such tokens alongside the ```textDocument/documentSymbol``` request... Even when the server declares supporting the capability... Maybe something needs to be done client side.
- I also tried, after seeing [this](https://github.com/microsoft/language-server-protocol/issues/1367), to send back a ```ServerCancelled``` error code when the document is busy parsing. The client receives the request but never sends a second one...

rtetley:
I will investigate some more and I am opening this PR anyways since it opens up important features:
- using custom error codes instead of always sending a RequestFailed one.
- giving a document_state (right now only Parsing and Parsed but later Executing and Executed maybe ?)

rtetley:
Even though this doesn't fix the original problem I think I will merge anyways because it adds interesting functionality. @gares ?

PR #960, opened 2024-12-06
Get proof blocks api

rtetley:
This PR aims at exposing an API which will allow extensions to query the document and get proof blocks in a structured view. This can be useful for a number of users who which to build an extension to vscoq.

Down the line we also ended up fixing an outline bug, in which symbols were never removed but always added.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -27,6 +27,8 @@ import GoalPanel from './panels/GoalPanel';
 import SearchViewProvider from './panels/SearchViewProvider';
 import {
     CoqLogMessage,
+    DocumentProofsRequest,
+    DocumentProofsResponse,
     ErrorAlertNotification,
     MoveCursorNotification, 
     ProofViewNotification, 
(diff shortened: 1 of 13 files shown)


language-server/protocol/extProtocol.ml
```diff
@@ -284,6 +301,9 @@ module Request = struct
     | "vscoq/documentState" ->
       let+ params = Lsp.Import.Json.message_params params DocumentStateParams.t_of_yojson in
       Pack (DocumentState params)
+    | "vscoq/documentProofs" ->
+      let+ params = Lsp.Import.Json.message_params params DocumentProofsParams.t_of_yojson in
+      Pack (DocumentProofs params)
```
gares, line 306: I'm wondering if we can easily version these custom queries, eg `"vscoq/documentProofs/v1"` so that we can update the format later without necessarily breaking badly the plugins that use this extension

client/src/extension.ts
```diff
@@ -59,6 +62,21 @@ export function activate(context: ExtensionContext) {
         });
     }
 
+    const getDocumentProofs = (uri: VersionedTextDocumentIdentifier) => {
+        const textDocument = TextDocumentIdentifier.create(uri.toString());
+        const params: DocumentProofsRequest = {textDocument};
+        const req = new RequestType<DocumentProofsRequest, DocumentProofsResponse, void>("vscoq/documentProofs");
+        Client.writeToVscoq2Channel("Getting proofs for: " + uri.toString());
+        client.sendRequest(req, params).then(
```
cas-haaijman, line 70: This promise should probably be returned

language-server/protocol/extProtocol.ml
rtetley: So after thinking about it, we make it accessible through a function in the client. If we need new versions it should be easy to keep everything wired on our end when plugins use our extension.

PR #962, opened 2024-12-10
Remove hypothesis display from goal list

rtetley:
If the goals are displayed as collabsibles, only the first goal has its goal context (hypothesis) displayed.

### DIFF
--- a/client/goal-view-ui/src/App.tsx
+++ b/client/goal-view-ui/src/App.tsx
@@ -1,7 +1,7 @@
 import React, {useState, useCallback, useEffect} from 'react';
 import "./App.css";
 
-import ProofViewPage from './components/templates/ProovViewPage';
+import ProofViewPage from './components/templates/ProofViewPage';
 import {Goal, ProofViewGoals, ProofViewGoalsKey, ProofViewMessage} from './types';
 
 import { vscode } from "./utilities/vscode";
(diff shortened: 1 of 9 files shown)


rtetley:
Screenshot 2024-12-10 at 14 54 10

Is this readable ? Should I also collapse them by default @ybertot ?

gares:
This is what we are used to in PG, CoqIDE & co.
About folding, I have no idea what is best. But the user choice should somehow stick.

TheoWinterhalter:
Will it be possible to show the hypotheses again if we want? Like maybe by unfolding it?

ybertot:
I thought I had answered one hour ago, but I would like to have the horizontal line displayed.

ybertot:
Well, no, I don't know why I thought it was the habit in proof-general, but it is not.  So as far as I am concerned, it is okay, but I think the question of Theo makes sense.

rtetley:
> Will it be possible to show the hypotheses again if we want? Like maybe by unfolding it?

I suppose I could put a collapsible for the hyps as well. Or add a button, but it will overload the UI a bit. Is it very useful ?

gares:
Is the tab view displaying the whole context? If so maybe Theo prefers it

TheoWinterhalter:
I don't actually care much for the tab display and I think the current proposal is what I would like in most cases so maybe it's not such a necessity.

ybertot:
If it is not too costly, I think a collapsible for the whole context would be good.

rtetley:
> If it is not too costly, I think a collapsible for the whole context would be good.

I need more context.... 
Pun aside what do you mean ? Like a nested collapsible for the hyps ? I can think of something

ybertot:
Yes, I meant : make the whole sequence hyps for each goal collapsible.

rtetley:
Screenshot 2024-12-11 at 10 53 42

This is what I came up with. What do you guys think ?

TheoWinterhalter:
Sounds great!

gares:
While you are at it, is it possible to avoid 
```
Goal 3
(3/3) ----
```
I mean, how many times do we want to tell the user the number of the goal? ;-)

rtetley:
> While you are at it, is it possible to avoid
> 
> ```
> Goal 3
> (3/3) ----
> ```
> 
> I mean, how many times do we want to tell the user the number of the goal? ;-)

As many times as possible it turns out x-D https://github.com/coq/vscoq/issues/674
Jokes aside, sometimes the context is very large and you don't see the "Goal 3" title after a scroll.

PR #963, opened 2024-12-11
Place the cursor at the end of the error when block on error mode is active

### LINKED ISSUE #952: enhancement: place cursor at error
In manual mode, when executing a command that raises an error, the location of the error is correctly highlighted in the faulty command, but the cursor is place at the end of the last correctly executed command.

The way syntax errors is handled is also awkward.  When trying to execute the next command, it does not fail if there is a syntax error (but it probably should), instead it tries to execute the next correctly formed command (which often out of context, or made of fragments that were not intended by the user).

rtetley:
Closes #952

### DIFF
--- a/language-server/dm/documentManager.ml
+++ b/language-server/dm/documentManager.ml
@@ -285,17 +285,14 @@ let state_before_error state error_id =
   match Document.get_sentence state.document error_id with
   | None -> state, None
   | Some { start } ->
+    let error_range = Document.range_of_id_with_blank_space state.document error_id in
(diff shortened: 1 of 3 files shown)


gares:
From the patch I can't tell if the error range is the range of the sentence (that you color with a red fading bg), of the error you underline.

I think the optimum would be the latter. Eg
```
Definition foo := not_existent 3.
```
should place my cursor at `|` and underline `^`:
```
Definition foo := not_existent| 3.
                  ^^^^^^^^^^^^
```

PR #974, opened 2025-01-08
Fix incorrect reference to vscoq.trace.server

cpitclaudel:
The README and the settings say to use `vscoq.trace.server`, but AFAICT this setting does nothing.  Looking at the code:

https://github.com/coq/vscoq/blob/26f3f1f70e329f72f1548f1a0ef929f20db36a4c/client/src/client.ts#L23-L28

… it appears that the right name is `vscoq-language-server`, not `vscoq`.  Indeed, with that setting set, I do get a trace (and I don't with the normal one).

An alternative fix would be to change the snippet above to have `vscoq` instead of `vscoq-language-server`.  Not sure what the consequences would be.


rtetley:
Thanks ! Took the liberty of rebasing, will merge once CI passes !

PR #978, opened 2025-01-14
Replaced custom searchForVscoqtopInPath code with which

cas-haaijman:
The client tests fail on my machine when running yarn test, but I tried installing a newly packaged .vsix and both my virtual windows and linux machines now find the correct path.

The error I am getting is:
DialogService: refused to show dialog in tests. Contents: Could not launch language server: ...
image


rtetley:
Thanks !

PR #1001, opened 2025-01-21
Fix messages in contiunous mode

### LINKED ISSUE #951: Inconsistent step highlighting
I am working with`vscoq-language-server 2.2.1` and `coq 8.19.2`
Given the code
```coq
Locate nat.

Goal True.
```
If I do an execution to point to the last line.
The first line does not get an executed status while it does if I do a step by step execution.
ddd

### LINKED ISSUE #997: Regression in continuous mode
- Messages no longer seems to work in continuous mode as was reported here: https://coq.zulipchat.com/#narrow/channel/237662-VsCoq-devs-.26-users/topic/Messages.20are.20not.20shown.20with.20updated.20VSCoq
- Probably related to #954 

This is probably due to the fact that commands that show feedback (such as ```Print```, ```Search```, etc...) are not stepped through when interpreting to position (https://github.com/coq/vscoq/issues/951)

rtetley:
This makes it so that queries and such are also stepped through when doing an interpret to position. This also fixes them in continuous mode.
Closes #997 
Closes #951

### DIFF
--- a/language-server/dm/scheduler.ml
+++ b/language-server/dm/scheduler.ml
@@ -200,7 +200,7 @@ let push_state id ast synterp classif st =
         base_id st, push_ex_sentence ex_sentence st, Exec ex_sentence
     end
   | VtQuery -> (* queries have no impact, we don't push them *)
-    base_id st, st, Query ex_sentence
+    base_id st, push_ex_sentence ex_sentence st, Query ex_sentence
   | VtProofStep _ ->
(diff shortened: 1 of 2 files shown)


rtetley:
@gares is this what you had in mind from #997 ?

gares (approved):

gares:
Yes

gares:
Yes

rtetley:
I'll fix the test suite and merge

PR #1002, opened 2025-01-21
Block on parse error

### LINKED ISSUE #877: Checking can "work" even if parsing fails
```
Goal forall {P }, P -> 0 = 0.
  contr 

Lemma foo : 0 = 0.
  intros.
  ```
  
image

### PR DESCRIPTION
**Pull request by rtetley**
At the moment the mechanism for parse errors are different than exec errors.
This is a limitation as we cannot block on the first parse error.
We introduce a mechanism to record a sentence state allowing us to detect that a sentence has a parse error and prevent from executing it.
Closes #877

### DIFF
--- a/language-server/dm/document.ml
+++ b/language-server/dm/document.ml
@@ -59,7 +59,7 @@ type comment = {
 type parsing_error = {
   start: int;
   stop: int;
-  msg: string Loc.located;
+  msg: Pp.t Loc.located;
   qf: Quickfix.t list option;
 }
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: language-server/dm/document.ml [hunk 1]
```diff
@@ -505,12 +531,15 @@ let get_entry ast =
 [%%endif]
 
 
-let handle_parse_error start msg qf ({stream; errors;} as parse_state) =
+let handle_parse_error start parsing_start msg qf ({stream; errors; parsed} as parse_state) =
   log @@ "handling parse error at " ^ string_of_int start;
   let stop = Stream.count stream in
   let parsing_error = { msg; start; stop; qf} in
+  let synterp_state = Vernacstate.Synterp.freeze () in
```
gares, line 538: I think you should just pass the synterp state you used to parser the sentence, rather than freezing it here.
I think in practice it will be the same, but there is no need to recompute the snapshot.

PR #1011, opened 2025-01-27
Fix parsing of command line arguments

### LINKED ISSUE #1003: Some vscoq.args are ignored?
It seems if I pass "-allow-rewrite-rules" through vscoq.args it is ignored, while having `-arg -allow-rewrite-rules` in a `_CoqProject` file works fine. I'm wondering if it is not linked to this: https://github.com/coq/coq/pull/19757/files#diff-596e8c750c790c5bcbf5425f7607226dc6742b60544601b3d5e876b329858bd4L408 @SkySkimmer @rtetley ?

### LINKED ISSUE #1007: missing backtrace
I'm experiencing crashes, but `-bt` does not print the trace, at least in this case:
```
[top                 , 908397, 1737677173.401560] ==========================================================
[top                 , 908397, 1737677173.401592] Trying to get range of non-existing sentence 1252
[top                 , 908397, 1737677173.401595] ==========================================================
```

It is hard to reproduce, I'm editing a large file and happens from time to time.

rtetley:
When we introduced the ability to read a _CoqProject on a per file basis, we also broke reading the command line arguments (which are typically given in the vscoq.args property from vscode).
This aims at fixing this. To avoid code duplication a new args module was introduced.

Closes #1003.
Closes #1007.

### DIFF
--- a/language-server/vscoqtop/dune
+++ b/language-server/vscoqtop/dune
@@ -1,7 +1,6 @@
 (executable
  (name vscoqtop)
  (public_name vscoqtop)
- (modules vscoqtop lspManager args)
  (flags -linkall)
  (package vscoq-language-server)
  (preprocess (pps ppx_optcomp -- -cookie "ppx_optcomp.env=env ~coq:(Defined \"%{coq:version.major}.%{coq:version.minor}\")"))
--- a/language-server/vscoqtop/dune
+++ b/language-server/vscoqtop/dune
@@ -1,7 +1,7 @@
 (executable
  (name vscoqtop)
  (public_name vscoqtop)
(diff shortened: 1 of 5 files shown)


language-server/vscoqtop/dune
```diff
@@ -1,7 +1,7 @@
 (executable
  (name vscoqtop)
  (public_name vscoqtop)
- (modules vscoqtop lspManager)
+ (modules vscoqtop lspManager args)
```
SkySkimmer, line 4: Why are you listing modules explicitly here instead of
```suggestion
```
?

SkySkimmer (approved):

Issue #989, opened 2025-01-19
vscoq is constantly crashing and has to be reloaded every minute


JasonGross:
I don't think the server should crash when it sees
```
[top                 , 91315, 1737325241.070620] ==========================================================
[top                 , 91315, 1737325241.070650] Trying to get range of non-existing sentence 794
[top                 , 91315, 1737325241.070651] ==========================================================
[Info  - 10:20:41 PM] Connection to server got closed. Server will restart.
true
[Error - 10:20:41 PM] Server process exited with code 0.
[top                 , 91390, 1737325246.720450] ==========================================================
[top                 , 91390, 1737325246.720488] Trying to get range of non-existing sentence 692
[top                 , 91390, 1737325246.720490] ==========================================================
[Error - 10:20:46 PM] Server process exited with code 0.
[Info  - 10:20:46 PM] Connection to server got closed. Server will restart.
true
[top                 , 91429, 1737325291.813120] ==========================================================
[top                 , 91429, 1737325291.813144] Trying to get range of non-existing sentence 1050
[top                 , 91429, 1737325291.813145] ==========================================================
[Error - 10:21:31 PM] Server process exited with code 0.
[Error - 10:21:31 PM] The Coq Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
```

rtetley:
Could you provide an example file where this happens ?

Dashadower:
I'm getting the same error. It's not occurring when I step through but when I'm in the middle of typing something.

JasonGross:
This was caused by having autocompletions enabled in settings, disabling them fixes the issue

Dashadower:
> This was caused by having autocompletions enabled in settings, disabling them fixes the issue

Did you disable vscoq's `Completion: Enable` option? I'm still getting the same error after disabling it, even after reinstalling coq and vscoq from a fresh switch.

rtetley:
I believe this is the same problem as #1009 and was fixed by https://github.com/coq/vscoq/pull/1021.
Next release is imminent !

Dashadower:
That's great. Looking forwards to it.

Blaisorblade:
This should probably be closed (fixed duplicate), unless the problem reoccurs.

gares:
Can you share a file where the problem occurs. Thanks

afdw:
Sure, here is one:
```rocq
Goal 1 = 1 -> 2 = 2.
  intros.
```

[...]

PR #1012, opened 2025-01-28
Return empty completions rather than error

Durbatuluk1701:
I believe this should address the spamming of `Request textDocument/completion failed` errors that can be encountered on each character if a completion is not found when `"vscoq.completion.enable": true`.

This does not address anything about the underlying completion algorithm, rather just ensures that we can (almost) always return an empty list of possible completions. 

I believe this is a more idiomatic way to implement completions (return either empty list or null) when no completions are available rather than a failure.


rtetley:
Thanks a lot ! Does this close #989 ?

Issue #248, opened 2021-08-24
Bad Indentation with `match ... end`


permui:
Expected Behavior: The `end` should be aligned with the pairing `match` after input.
Actual Behavior: `end` goes to the outmost indentation.
Illustration: 
![2021-08-24-11-16-03](https://user-images.githubusercontent.com/34205956/130550639-4b898cf5-cd72-4213-bf22-b8b817d8ee2c.gif)
This is quite annoying and I could not find any settings related to this.
Hoping a fix. Thanks very much.

rtetley:
Related to #247 and #158

shilangyu:
Without a full blown formatter, I think what was done in #1015 is the best that can be done regarding this issue.

I had a look at what rust-analyzer does (notoriously well-developed LSP for Rust), it also does not handle this correctly.

Durbatuluk1701:
> Without a full blown formatter, I think what was done in #1015 is the best that can be done regarding this issue.

I tend to agree, I think #158 is probably the optimal tracking issue (for a hopefully eventual formatter)

shilangyu:
@gares I am closing as I believe we have implemented the best version possible. To improve it we would need a full code formatter.

Issue #247, opened 2021-07-15
Incorrect Indentation with `Show Proof.`


benrbray:
When typing `Show Proof.` followed by ENTER, the new line will have a higher indentation level, as if a new proof has been started.  Similarly, when typing `Qed.` in a proof that contains `Show Proof.`, the `Qed.` will be automatically indented to the same level as `Show Proof.`, rather than to the level of the proof itself.

![vscoq-showproof](https://user-images.githubusercontent.com/221963/125818574-1b1bc619-953e-49ea-994d-23949bee4b68.gif)

Is this within the scope of vscoq or does it require a change to the language server?  Thanks!

rtetley:
Related to #158

shilangyu:
This can be closed because it was fixed by @Durbatuluk1701 in #1015

Issue #158, opened 2020-09-03
Format Document doesn't work


ogiekako:
Reproduction:
1. Open foo.v, which is not formatted, e.g.
```
Inductive day : Type :=
  | monday
    | tuesday
  | wednesday
    | thursday
  | friday
  | saturday
  | sunday.
```
2. Press F1 && Format Document.

Got:
- The source code isn't formatted.
- VS Code show the message "There is no formater for 'coq' files installed". 
- I press the button "Install Formatter..." on the message, but no candidates are listed.

Expected:
The source code is formatted.

Zimmi48:
But there is no formatter for Coq source code, so I don't see why you would expect a different behavior. Furthermore, this is not really related to VsCoq, is it? The "Format Document" command is provided by default by VsCode.

gares:
The I guess this bug can be turned into a feature request...

Zimmi48:
There are good chances though that the beautifier will not be a satisfying formatter (at least not before many fixes). It has hardly been maintained at all.

shilangyu:
I really don't think this feature is vsrocq's concern. At best, if a formatter exists, vsrocq could integrate it. But definitely not implement it.

PR #1015, opened 2025-01-28
Change Language Config to Fix Indents

Durbatuluk1701:
Couple Changes Here:
- Change the language configuration files to `.language-configuration.json` (as suggested by [lang guide](https://code.visualstudio.com/api/language-extensions/language-configuration-guide))
- Modify the language configuration to address #247 (manual proof indentation with negative lookbehind)
- Modify the `wordPattern` definition to correctly identify identifiers in Rocq (items with a `'` within them before would not be correctly suggested by intellisense)
- Added some auto-indent features for Sections, Module Types, and Modules.

IndentFixesGif

Of Note:
- Changes were made to make it so that match statements do not auto-indent (related to #248), but I am not sure the solution is optimal? I am seeking feedback if we think the behavior exhibited below is better or worse than the current approach?
MatchIndentChanges


Durbatuluk1701:
I still believe #158 would be helpful (and could work on addressing it in the future), but I think hopefully these changes would make some of the default indentation that occurs while typing better

rtetley:
I believe this is already an improvement so I'll merge ! Thanks !

Issue #985, opened 2025-01-16
Feature: change color of messages


FissoreD:
Hello,
not sure if the option already exists, but I use the light theme in vscode and 
I find it difficult to read the messages (shown in yellow).
Is it possible to change this color ?

Here is a screenshot of what I'm seeing.
![Image](https://github.com/user-attachments/assets/7ee9efa1-bc5d-4aca-b89e-03bceceb2088)

4ever2:
You can overwrite the colors that the extension uses by adding the following to your settings:
```json
"workbench.colorCustomizations": {
    "coq.warningMessage":"#ffff00"
}
```

The names and default values of the colors can be found here https://github.com/coq/vscoq/blob/34ead82bdb2c07150acfa664117358ed03c63620/client/package.json#L633

FissoreD:
Oh, too good.
Thank you!

TheoWinterhalter:
Maybe the issue should stay open if it happens with every/many light theme?

rtetley:
I will tweak the color them. Pretty sure I can set up a dark/light thing.

rtetley:
Closed by #1019

PR #1019, opened 2025-01-29
Set light and contrast theme colors for warning messages to amber

Durbatuluk1701:
This should hopefully address #985 

How it appears for one of the default vscode light themes:
image

How it appears for one of the default vscode dark themes:
image


Durbatuluk1701:
Not sure what went wrong in the CI, maybe missing the libsecret dependency on certain runners? 

Seems related: https://github.com/atom/node-keytar/issues/345

PR #1023, opened 2025-01-31
Adapt w.r.t coq/coq#20167.

ppedrot:
This is backwards compatible.

I copied the function here because Tok.equal in Coq is actually not the equality function, there is special-casing of KEYWORD and IDENT that used to be there to support a hack in the parser.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,7 +4,7 @@
   inputs = {
     flake-utils.url = "github:numtide/flake-utils";
 
-    coq-master = { url = "github:coq/coq/d6dd8a618cbf628a6580c9a42db0cf5f44606bea"; }; # Should be kept in sync with PIN_COQ in CI workflow
+    coq-master = { url = "github:coq/coq/3e7c26f8bb991e7335162647017edf21b2afd974"; }; # Should be kept in sync with PIN_COQ in CI workflow
     coq-master.inputs.nixpkgs.follows = "nixpkgs";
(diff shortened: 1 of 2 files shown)


ppedrot:
I'm not sure why the CI is still broken here, but it should really be backwards compatible, I tested it locally.

SkySkimmer:
The CI for master seems to have issues with the stdlib split or renaming 
> Error: I don't know about package coq-stdlib (passed through --only-packages)

CI for <9.0 correctly says 
>Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
(PATTERNIDENT _, _)

gares:
What is Tok.equal then, and why it is called like that?

ppedrot:
@gares it was introduced for backward compatibility with camlp4 quite a long time ago, and this IDENT ≤ KEYWORD rule was added to work around the fact that an indent could be turned at any point into a keyword. So that's a plain hack.

ppedrot:
Anyone merging this?

SkySkimmer:
This branch has conflicts that must be resolved

ppedrot:
Rebased.

SkySkimmer:
You still have the non exhaustive match issue on older Coq versions

ppedrot:
Hopefully this does not trigger warnings or something.

Issue #1027, opened 2025-02-07
VSCoq complains about uninstalled server, even outside of Coq files


TheoWinterhalter:
I have some opam switches without the server because they don't even have Coq installed, but whenever I use them, VSCoq complains that I should install it. I believe it should only do so if it detects I'm trying to do some Coq.

I don't know how hard or reasonable it is, but it feels like it shouldn't even run in those cases.

Durbatuluk1701:
I believe VsCoq attempts to "activate" whenever it detects a `.v` or `_CoqProject` file in the workspace. Do you believe it is activating even when none of those are present? Or are you saying you want it to only activate once a `.v` or `_CoqProject` file has been opened?

TheoWinterhalter:
Ah, you're right, I was working on some F* but there is *one* .v file somewhere. Maybe it should wait for those to be opened yes?

Also, it wouldn't be as annoying if it weren't a modal window and just a warning notice in the bottom right.

Durbatuluk1701:
I would have no objections to the activation waiting until you open a Coq related file, but maybe others have some workflow where they start utilizing the VsCoq functionality (like the Search panel in the sidebar maybe?) before opening one?

TheoWinterhalter:
Maybe, but as far as I remember, the Search panel doesn't actually work unless you've loaded a file.

Durbatuluk1701:
Looks like that is correct, seems like the reasons not to do this are few and far between, if not non-existent

PR #1028, opened 2025-02-07
Change activation to only occur when Coq files are opened

Durbatuluk1701:
Related to #1027.
In particular, you will not need the language server involved unless the extension is "activated", and **now** it should only activate if you have a focused `.v` or `_CoqProject` file.

How this should work (and in some ways a continuation of discussion in #897):
| Condition | Activate VsCoq |
| ----------- | ----------------------- |
| *.v file exists in the workspace | False |
| _CoqProject file exists in the workspace | False |
| *.v file is open | False |
| _CoqProject file is open | False |
| *.v file is focused | **True** | 
| _CoqProject file is focused | **True** |

It may also be worth addressing in this PR suggestions made by @RalfJung (in #897 discussion) about not displaying the sidebar unless you have a currently focused `.v` file. Upon testing, it appears that the sidebar search will not work if you don't have a currently focused `.v` file, so displaying it when it won't work would be at best confusing. 
If we choose to implement this, then additional commits will be needed for that functionality

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -29,10 +29,7 @@
   "readme": "https://github.com/coq-community/vscoq/blob/master/README.md",
   "bugs": "https://github.com/coq-community/vscoq/issues",
   "homepage": "https://github.com/coq-community/vscoq/blob/master/README.md",
-  "activationEvents": [
-    "workspaceContains:**/_CoqProject",
-    "workspaceContains:**/*.v"
-  ],
+  "activationEvents": [],
   "main": "./dist/extension.js",
   "contributes": {
     "languages": [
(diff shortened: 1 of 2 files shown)


rtetley:
I think displaying the sidebar only when a .v file is focused sounds fine !
Rebasing to main should fix CI !

Durbatuluk1701:
@rtetley any ideas on why the CI is failing now? It seems odd that it fails for macos 8.18 and 8.20, but succeeds for 8.19?

rtetley:
Probably sporadic fails ! I'll re-run !

PR #1034, opened 2025-02-12
log: safe generation of messages

gares:
This avoids crashes if the log message printer fails


gares:
Looks green now.

gares:
Given the number of files this patch touches, I think it should be merged ASAP or it will likely conflict

PR #1039, opened 2025-02-19
ci: add a retry action to avoid sporadic fails

rtetley:


rtetley:
@gares this is bound to fail as proof delegation is currently broken. Nonetheless could you check at the test files in testFixtures and make sure these are the intended behaviours ? The fact that we should have diagnostics when we skip proofs is suspicious to me !

PR #1045, opened 2025-02-19
Symbol search: handle inductive, fixpoint, and cofixpoint

KacperFKorban:
Previously, doing symbol search in a file didn't show some definition and declaration types, specifically `Fixpoint`, `CoFixpoint`, and `Inductive`. Now they should be indexed and displayed correctly.

### DIFF
--- a/language-server/dm/document.ml
+++ b/language-server/dm/document.ml
@@ -190,6 +190,9 @@ let record_outline document id (ast : Synterp.vernac_control_entry) classif (out
         match pure with
         | Vernacexpr.VernacStartTheoremProof (kind, _) -> Some (TheoremKind kind)
         | Vernacexpr.VernacDefinition ((_, def), _, _) -> Some (DefinitionType def)
+        | Vernacexpr.VernacInductive (kind, _) -> Some (InductiveType kind)
+        | Vernacexpr.VernacFixpoint (_, _) -> Some (DefinitionType Decls.Fixpoint)
+        | Vernacexpr.VernacCoFixpoint (_, _) -> Some (DefinitionType Decls.CoFixpoint)
         | _ -> None
     in
     let name = match names with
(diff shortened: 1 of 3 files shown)


language-server/dm/document.ml
```diff
@@ -189,6 +190,9 @@ let record_outline document id (ast : Synterp.vernac_control_entry) classif (out
         match pure with
         | Vernacexpr.VernacStartTheoremProof (kind, _) -> Some (TheoremKind kind)
         | Vernacexpr.VernacDefinition ((_, def), _, _) -> Some (DefinitionType def)
+        | Vernacexpr.VernacInductive (kind, _) -> Some (InductiveType kind)
+        | Vernacexpr.VernacFixpoint (_, _) -> Some (DefinitionType Decls.Fixpoint)
+        | Vernacexpr.VernacCoFixpoint (_, _) -> Some (DefinitionType Decls.CoFixpoint)
```
gares, line 195: Are you sure these are really needed?

gares (commented):
Many thanks for your contribution

language-server/dm/document.ml
KacperFKorban: I wasn't sure, I added it to cover all my bases. I tested it now, and from what I can tell the fix works without this change, so I removed this part of the diff.

gares (approved):

gares:
CI is morally green (waiting for #1039)

language-server/dm/document.ml
gares: Thanks @SkySkimmer

SkySkimmer: you don't need the Inductive case here since there's no such thing as an interactive Inductive

rtetley:
Thanks !

PR #1046, opened 2025-02-21
[feat] make Rocq interruptible

gares:
TLDR: Rocq's interpreter runs in a thread that can be interrupted.

Screenshot From 2026-04-01 13-04-33
Screenshot From 2026-04-01 13-06-11

### Technology

Requires sel 0.7 (for promises) and memprof-limits (battle tested in coq-lsp, kudos to @gadmm)

### Expected behavior
- in stop-on-first-error-mode it just stops
- in continue-on-error it continues (maybe this is not the best behavior, but this is how it is in this PR)

Editing the document interrupts execution as usual, but now also if the Rocq interpreter is looping. Hence I don't think there is a real need for a keybinding to interrupt Rocq. Thanks to this, continuous mode is usable if looping sentences are around.

### Other design choices

The proverThread has 3 ways to sequentialize:
- eventually_run: put in a queue and return a promise
- try_run: check if the thread is busy, and if not run now (with a timeout)
- run: if the thread is busy, interrupt and put in the queue the interrupted task, then run now (with a timeout)

This PR uses *eventually_run* for parsing and execution (checking) of sentences; *try_run* for queries (including hovering); *run* for he rest, including printing the proof view. Note that you can check two documents at the same time, vsrocq always did "time sharing" bu checking one sentence of each, so busy really means "busy on any open document". These choices are trivial to change in the code. The main tradeoff is that interrupting/restarting jobs loses time. If, for example, hovering does that, one may really slow down a long sentence (each time you move the mouse you risk to restart it).
 
## Architecture

There was a big issue, namely that Rocq is not thread safe (nor purely functional). Scenario: the execution manager (thread) is running on a sentence: it installed a global state and is possibly even modifying it in place before snapshotting it back; at the same time the lsp manager gets a request, like hover or query or completion, that currently installs a Rocq state and run some code.  Hence we reorganize the code as follows

Screenshot From 2026-04-01 12-57-30

All calls to Rocq APIs are either pure, or serialized by the prover thread. Queries are run on a best effort (if the thread is busy for too long they are aborted). Parsing/Execution is serialized and made interruptible by the user.
A test compares the call graph of documentManager with a reference, checking that each call is either whitelisted (pure code) or goes trough a safety barrier (makes the call sequential).

- [x] block on first error broken


gares:
CC @rtetley, when you have a minute, could you please add a stop button to the UI ?

gares:
After some thinking the plan could be:
- rework the source tree to use `(implicit_transitive_deps false)` in dune project and have all Rocq calls / types in a dedicated library `rocq_enclave`. All code outside the enclave should not list coq-core.* in the dune file, hence it cannot call Rocq APIs. In this way we precisely control what we call
- make sure that Rocq calls are made in a single thread and serialized one after the other
  - with an API to interrupt, possibly using memprof_limits
  - use the same priorities we use for sel in the API: if a call has higher priority interrupt the running one
  - bonus: if a call has higher priority interrupt the running one (after say 0.1) and run the high prio one, then restart the interrupted one

This includes all Rocq calls, the queries, the parser, etc

Some Rocq code may not need this, being pure, but it is not so clear how to know it and ensure it Rocq side. Eg `Search` seems pure, but prints the results using the current notations, so it is not really pure.

rtetley:
> CC @rtetley, when you have a minute, could you please add a stop button to the UI ?

Done



> Along with a stop button could we have a key binding to interrupt as well

I created a command that you can call from the palette. What default keybinding should we use ?

Durbatuluk1701:
I know it has been previously suggested that if you have an in-progress proof and input the "Step Backward" key bind it should be interpreted as "Interrupt", so that is a possibility.

I also think maybe given that we have commands bound to up, down, and right arrow keys, maybe left arrow key could be "Interrupt"?

language-server/protocol/extProtocol.ml
```diff
@@ -76,6 +85,9 @@ module Notification = struct
       | "prover/interpretToEnd" ->
         let+ params = Lsp.Import.Json.message_params params InterpretToEndParams.t_of_yojson in
         InterpretToEnd params
+      | "vscoq/interrupt" ->
```
gares: prover!

client/src/manualChecking.ts
```diff
@@ -1,19 +1,15 @@
...
-} from 'vscode-languageclient/node';
-
-import GoalPanel from './panels/GoalPanel';
-
 import Client from './client';
 import { makeVersionedDocumentId } from './utilities/utils';
 
+export const sendInterrupt = (editor: TextEditor, client: Client) => {
+    const textDocument = makeVersionedDocumentId(editor);
+    client.sendNotification("vscoq/interrupt", {textDocument});
```
gares: prover

gares: document extension

client/package.json
```diff
@@ -128,14 +128,20 @@
...
             "scope": "workspace",
             "type": "number",
             "default": 4,
             "description": "Sets the memory limit over which we discard the execution state of documents which have been saved (in Gb). "
+          },
+          "vsrocq.interrupt.preempt": {
+            "scope": "workspace",
+            "type": "boolean",
+            "default": false,
+            "description": "Hovering and other queries preempt checking tasks (recommended Rocq >= 9.2). "
```
gares:
```suggestion
            "description": "Hovering and other queries preempt checking (recommended Rocq >= 9.2). "
```

PR #1047, opened 2025-02-24
syntax highlighting: support more vernacular commands

KacperFKorban:
(Some more commands/command combinations, that I noticed, while doing my coq proofs)

Highlight the following vernacular commands/combinations:
- Hint Mode
- Create HintDb
- Print All

Also, refactor the "Vernacular keywords" regex.


rtetley:
Thanks !

PR #1053, opened 2025-02-26
wip: move rocq API behind an abstraction

gares:
objective: track stateful APIs so that we can eventually run them in an interruptible thread (one at a time)

this is preliminary work for #1046


gares:
This approach is not going to work, I should proceed into smaller steps

Issue #137, opened 2020-05-23
Allow folding of Proofs and Sections


Arkhist:
VSCoq should allow its users to fold Proof/Qed blocks and Section/End blocks.

I believe this feature should be implemented on the language-server side: implementing it as a language configuration feature could create conflicts between the different block types.

swils:
I'm only learning Coq, so not an expert, but I'm not convinced that this needs something on the side of the language server. In my personal experience, VSCode already folds a lot of stuff properly, at least for the simple proofs that I'm currently doing. Could you perhaps send me a couple of snippets that you think cannot be folded properly without the language server?

(By coincidence, the last few days I have been looking at the internals of VSCoq. And perhaps changes to the TextMate definition might allow better folding. Currently everything is in the patterns section; but not much in the repository section. I think (but not sure) that improving this might help with folding. And an unrelated thing that could potentially also be improved is the binding to the scope selectors, spreading it out over more TextMate root groups.)

Arkhist:
I inspected the source code of the extension and it does seem like all of the folding happening is caused by the auto folding of VSCode itself when using proper indentation.

I am working with libraries that don't use indentation in their proofs and in files with more than 2000 lines, folding becomes very useful.

Here is a minimal example where folding doesn't work.
```Coq
Section test.

Theorem foo: forall a b: Prop, a /\ b -> b /\ a.
Proof.
intros.
split.
destruct H as [Ha Hb].
exact Hb.
destruct H as [Ha Hb].
exact Ha.
Qed.

Theorem bar: forall a b: Prop, a \/ b -> b \/ a.
Proof.
intros.
destruct H.
right.
assumption.
left.
assumption.
Qed.

End test.
```

TheoWinterhalter:
Note that `Proof` is not mandatory at the beginning of a proof, which might make folding harder.

swils:
No @Arkhist you're right. Improving the TextMate file does not improve the folding. Too bad...

rtetley:
Related to #799

PR #1059, opened 2025-03-03
feat: adding sections and modules in outlines

rtetley:
We now support sections and modules which contain document symbols in the outline.
This is a preliminary step for #137

### DIFF
--- a/language-server/dm/documentManager.ml
+++ b/language-server/dm/documentManager.ml
@@ -351,6 +351,18 @@ let get_document_proofs st =
   let proofs, _  = List.partition is_theorem outline in
   List.map mk_proof_block proofs
 
+let to_document_symbol elem =
+  let Document.{name; statement; range; type_} = elem in
+  let kind = begin match type_ with
+  | TheoremKind -> SymbolKind.Function
+  | DefinitionType -> SymbolKind.Variable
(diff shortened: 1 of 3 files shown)


rtetley:
I need to make the ```sec_or_m``` variable into a queue in order to support imbricated sections / modules

language-server/dm/documentManager.ml
```diff
@@ -351,19 +351,55 @@ let get_document_proofs st =
...
+  in
   let to_document_symbol elem =
     let Document.{name; statement; range; type_} = elem in
     let kind = begin match type_ with
     | TheoremKind _ -> SymbolKind.Function
     | DefinitionType _ -> SymbolKind.Variable
     | InductiveType _ -> SymbolKind.Struct
     | Other -> SymbolKind.Null
+    | BeginSection | BeginModule -> SymbolKind.Class
+    | End -> SymbolKind.Null
```
gares, line 376: This mapping also seems to belong to the protocol/ folder, IMHO.

rtetley:
I removed the un-necessary info, however I don't think the types should go to protocol. They have nothing to do with the protocol itself. Not sure where to put the mapping though.

rtetley:
I'll merge this, the moving will happen when I start re-architecturing the code.

PR #1063, opened 2025-03-10
fix: remember to compare errored sentences in the diff

rtetley:


language-server/dm/document.ml
```diff
@@ -447,9 +447,14 @@ let tok_equal t1 t2 =
   | QUOTATION(s1,t1), QUOTATION(s2,t2) -> CString.equal s1 s2 && CString.equal t1 t2
   | _ -> false
 
+let same_errors (e1 : parsing_error) (e2 : parsing_error) =
+  let s1 = Pp.string_of_ppcmds (snd e1.msg) in
+  let s2 = Pp.string_of_ppcmds (snd e2.msg) in
+  (s1 = s2) && (e1.start = e2.start) && (e1.stop = e2.stop)
```
gares, line 453: aren't the start/stop possibly shifted? Also I think the tokens can be compared as well (although the tokenizer can raise an error).

If this check is just an approximation, it should be documented I guess.

PR #1067, opened 2025-03-14
Ignore nested proofs option instead of failing with anomaly

### LINKED ISSUE #1060: vscoq gives anomalies on attempting to `Require` a file containing `Global Set Nested Proofs Allowed.`
Image

sorry, not sure how to reproduce this one

SkySkimmer:
when it is set in a .vo we Require.
fix #1060


gares:
Thanks!

PR #1078, opened 2025-04-08
Rename extension

rtetley:


client/src/extension.ts
```diff
@@ -98,18 +94,18 @@ export function activate(context: ExtensionContext) {
...
     );
     
-    // Detect if vscoq1 is installed and active
-    const vscoq1 = extensions.getExtension("coq-community.vscoq1");
-    if (vscoq1) {
-        if (vscoq1.isActive) {
-            const message = "VsCoq2 is incompatible with VsCoq1. it is recommended that you disable one of them.";
-            window.showErrorMessage(message, { title: "Disable VsCoq1", id: 0 }, { title: "Disable VsCoq2", id: 1 })
+    // Detect if vsrocq1 is installed and active
+    const vsrocq1 = extensions.getExtension("rocq-community.vsrocq1");
```
gares, line 98: No sure

client/src/manualChecking.ts
```diff
@@ -17,21 +17,21 @@ import { makeVersionedDocumentId } from './utilities/utils';
 export const sendInterpretToPoint = (editor: TextEditor, client: Client) => {
     const textDocument = makeVersionedDocumentId(editor);
     const position = editor.selection.active;
-    client.sendNotification("vscoq/interpretToPoint", {textDocument: textDocument, position: position });
+    client.sendNotification("vsrocq/interpretToPoint", {textDocument: textDocument, position: position });
```
gares, line 20: This breaks the protocol, it must be made super clear this is a one time thing...

client/src/test/runTest.ts
```diff
@@ -2,7 +2,7 @@ import * as path from 'path';
 import * as fs from 'node:fs/promises';
 import * as tmp from 'tmp-promise';
 
-import { runTests, downloadAndUnzipVSCode } from '@vscode/test-electron';
+import { runTests, /* downloadAndUnzipVSCode */ } from '@vscode/test-electron';
```
gares, line 5: ?

gares (approved):
Amazing and scary

gares:
I wonder if, while er are at it, we should not remove Rocq from some identifiers. It is everywhere

client/src/extension.ts
rtetley: Ooops... WIll fix tomorrow...

client/src/manualChecking.ts
rtetley: Yeah... I think in the line of the comment below, maybe we should remove rocq from these verbs...

client/src/test/runTest.ts
rtetley: ah yeah not sure this is necessary indeed..., I think I kept it because of a comment below, which I suppose we could remove as well !

PR #1082, opened 2025-04-19
Fix types of values sent to marshal

SkySkimmer:
We only unmarshal (`Sel.On.ocaml_value`) at types Job.t (in setup_plumbing initialization) and `worker_message`, so it does not make sense to marshal other types.

In particular, install_feedback_worker and the send_back produced by setup_plumbing seem like they were incorrect.

It also seems like nothing is testing these code paths since the tests pass both before and after this change. We should try to find some tests which fail in master and pass with this before merging.

### DIFF
--- a/language-server/dm/delegationManager.ml
+++ b/language-server/dm/delegationManager.ml
@@ -23,7 +23,7 @@ type link = {
   read_from:  Unix.file_descr;
 }
 
-let write_value { write_to; _ } x =
+let write_value_gen { write_to; _ } x =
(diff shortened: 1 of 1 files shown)


gares (approved):

gares:
Thanks

SkySkimmer:
>We should try to find some tests which fail in master and pass with this before merging.


...

rtetley:
Oops sorry, a little tired that slipped my mind... No worries I'll fix it

SkySkimmer:
I guess we can still checkout the older commit for testing so it doesn't really matter

PR #1087, opened 2025-05-08
Use all names from a vernac gen expr, when recording outline

KacperFKorban:
Create entries in the outline for all symbols in a mutually recursive fixpoint definition.

image


rtetley:
Thanks ! The CI should be fixed if you rebase :-)

PR #1088, opened 2025-05-08
When moving to top end update_view and mk_proof_view events

KacperFKorban:
closes coq/vscoq#1084


rtetley:
Thanks again :) same ! If you re-base I should be able to merge !

PR #1089, opened 2025-05-09
Add syntax highlighting for Ltac2 term definitions

KacperFKorban:


rtetley:
Same as other PRs :) re-basing should fix !

PR #1107, opened 2025-05-22
Add match! keywords for Ltac2 to syntax highligthing

thomas-lamiaux:


Durbatuluk1701:
Might we want to modify the `rocq.tmLanguage.json` too?
https://github.com/rocq-prover/vsrocq/blob/98927e92c4c83cee1901d1e9bba6fcadbdf4457f/client/syntax/rocq.tmLanguage.json#L149

rtetley:
Thanks !

PR #1114, opened 2025-05-23
Fix vsrocqtop usage display

vincent-botbol:
This PR fixes an issue where the `vsrocqtop_specific_usage` would be evaluated too early and would miss subsequent logger registration. We fix it by making it a thunk.


vincent-botbol:
Before:
image

After:
image

gares (approved):

PR #1133, opened 2025-06-27
feat: add an option to render goal view as formatted string + add range

rtetley:
Now the server can be configured to return the formatted string instead of the pp. Additionally we return the range of the sentence corresponding to the rendered goal view.


rtetley:
@cpitclaudel ?

rtetley:
It's better with all the files :rolling_eyes:

rtetley:
Hmmm I probably broke the protocol on the front end side. I will fix it later !

PR #1134, opened 2025-07-10
Fix 1066

### LINKED ISSUE #1066: Bad syntax error recovery
I am with vscoq  2.2.5 coq 8.20.1

Ok I have a file with an obvious syntax error. Anyway I try to execute it anyway (this is important)

Image

now if I try to correct the error, the error message remains shitted of some characters.

Image

If I don't try to execute the line with a syntax error, correcting works fine.

gares:
@rtetley See my last comment on the issue. I noticed that some piece of code does install the wrong synterp state. My guess is that if a "parse error" sentence gets executed it is promoted to a "exec error" sentence with a full (as in interp+synterp) state. That state, for reasons I could not fully undertand is the used to reparse the sentence, or is installed by someone else and the just re-installing the synterp state is not enough...).

I think my fix is just a safety-belt, one would need to understand more that causes the wrong state to be used for parsing.
I suspect also Skip sentences suffer the same.

Fix #1066

### DIFF
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -78,7 +78,7 @@ type delegated_task = {
 
 type prepared_task =
   | PSkip of { id: sentence_id; error: Pp.t option }
-  | PBlock of { id: sentence_id; error: Pp.t Loc.located }
+  | PBlock of { id: sentence_id; synterp : Vernacstate.Synterp.t; error: Pp.t Loc.located }
   | PExec of executable_sentence
   | PQuery of executable_sentence
   | PDelegate of delegated_task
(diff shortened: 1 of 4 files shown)


language-server/dm/executionManager.ml
```diff
@@ -616,7 +616,8 @@ let execute_task st (vs, events, interrupted) task =
   end else
     try
       match task with
-      | PBlock { id; error = err} ->
+      | PBlock { id; synterp; error = err} ->
+        let vs = { vs with Vernacstate.synterp } in
```
gares, line 620: this is the important line

PR #1149, opened 2025-09-24
Update vscoq-language-server.opam

patrick-nicodemus:
I think this is right?


TheoWinterhalter:
I think the point is to rename to Rocq not the other way around?

PR #1150, opened 2025-09-26
Add basic MCP support on the client side

KacperFKorban:
## Summary

This PR makes the following changes:
1. Adds an MCP server with basic ways for progressing/checking the state of the proof. It includes tools for step*, interpret* and proofView actions. (There is a new setting to enable the MCP server, and an additional setting to set the MCP server port)
2. Changes step* and interpret* events from notifications to requests. (This is done, because LSP requests have their unique ids, when notifications don't). Thanks to this change, the proofView event can also have an optional request_id that identifies the request that triggered this notification.

## Discussion

> Why add the MCP server/support on the client side?
We initially discussed putting the MCP support on the language-server side, but after going back and forth, I decided to go back to the first approach and do it as part of the vscode extension. The main reasoning was:
1. Both the server and the client work in an asynchronous way, so there will still be quite a bit of mess related to waiting for the LSP response, and I'd rather have this mess be on the extension side, where it can be reliably decoupled.
2. MCP servers are almost exclusively used with vscode-clones, so it doesn't introduce any limitations.


gares:
Thanks!

PR #1151, opened 2025-09-29
Moved hardcoded colors to themed colors in package.json

simonjwinwood:
Previously Decorations.ts had hardcoded colors for the various status indicators (background, rule, outline) for the prepared/processing/processed regions.  This commit moves these colors into package.json so that users can customise them.

New colors are:
 - rocq.prepared.ruler
 - rocq.prepared.outline
 - rocq.prepared.background
 - rocq.processing.ruler
 - rocq.processing.background
 - rocq.processed.ruler
 - rocq.processed.background

The defaults were set to be those from Decorations.ts where possible.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -964,7 +964,7 @@
         "id": "rocq.processed.background",
         "description": "Bazckground color of processed region.",
         "defaults": {
-          "dark": "#01100133",
+          "dark": "#00960033",
           "light": "#00960033",
           "highContrast": "#01100133"
         }


gares:
I'm not an expert here, so I just tested it. Without further tweaking there is a regression.

Before:
Screenshot from 2025-10-23 10-05-05
Screenshot from 2025-10-23 10-04-57

After:
Screenshot from 2025-10-23 10-05-44
Screenshot from 2025-10-23 10-05-24

I just tested the default light/dark modes, and the default color for "processed" for the dark mode is just invisible

PR #1157, opened 2025-10-19
Nix build fixes

mbrcknl:
Note that this PR targets the CI-9.1 branch (#1156), since it includes some changes relative to that branch.

**Add preBuild to make dune-files**

The language-server opam build was previously updated to perform `make dune-files` before running `dune build`. This commit makes the same change in the Nix-based build. Without this change, only the version parser is built.

**Fix language server builds for 9.0/9.1**

- Use `coq_9_*` instead of `rocq-core`, because Dune requires `coqc`. Also use explicit versions.
- Use nixpkgs-unstable for 9.1, because it is not found in nixos-25.05.

**Remove unused buildInputs**

The Nix-based language-server builds previously specified several GTK-related build inputs. These do not seem to be relevant to the language server, and might have been inadvertently copied from the RocqIDE build. This commit removes them.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -50,10 +50,6 @@
                   dune_3
                 ]
                 ++ (with coq.ocamlPackages; [
-                  lablgtk3-sourceview3
-                  glib
-                  pkgs.adwaita-icon-theme
-                  wrapGAppsHook
                   ocaml
                   findlib
                   yojson
@@ -96,10 +92,6 @@
                   dune_3
                 ]
                 ++ (with coq.ocamlPackages; [
-                  lablgtk3-sourceview3
-                  glib
-                  pkgs.adwaita-icon-theme
-                  wrapGAppsHook
                   ocaml
                   yojson
                   findlib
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -4,6 +4,7 @@
   inputs = {
     flake-utils.url = "github:numtide/flake-utils";
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.05";
+    nixpkgs-unstable.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
```
mbrcknl: I've added nixpkgs-unstable as a separate input here, to minimise differences to the previous configuration. However, I would recommend using nixpkgs-unstable for everything, optionally pinning a commit, so that there is only one copy of nixpkgs involved.

Let me know if you'd like me to change this.

mbrcknl:
The dev-setup-opam failures seem unrelated to anything in this PR or in #1156. I suspect dev-setup-opam might also currently fail on main.

PR #1165, opened 2025-10-24
Fix minor typos/stale links in github issue templates

JoJoDeveloping:
It's missing an `s`. Otherwise people try to report an issue, only to have their language server immediately crash.


gares:
Thanks

PR #1172, opened 2025-11-04
dm: do not crash on outdated SendProofView event

gares:
This bug was added in https://github.com/rocq-prover/vsrocq/pull/1133 but I think it raises a general concern.
LSP events have high priority, then can arrive while the document is being parsed. In response to these one may generate a lower priority event, eg exec_to or update_pv. If these events contain sentences ids, this is problematic because they refer to an old document, i.e. when parsing ends they may become non-existent.

This patch makes the code of SendProofView resilient.
The code of ExecTo is also resilient, there is a comment saying "should we error?".

In general it is not so clear if the best solution is to write resilient code or making the events never contain potentially stale info. The first option, as in this PR, makes some user action ignored. When I reproduce the bug I see my "goto next" not being run (this was already the case) and the PV not being updated (used to crash). The second option complicated the code a little, i.e. an lsp event (top prio) does not generate an exec_to or update_pv event (lower prio) looking at the document. It generates a request_exec_to event (lower prio) containing the caret position (not the sentence id or tasks), and that even is picked up only when higher priority events terminate, like parsing that updates the document.

WDYT @rtetley

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -6,7 +6,7 @@
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-25.05";
     nixpkgs-unstable.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
 
-    rocq-master = { url = "github:rocq-prover/rocq/aeabefe987b1a9aa6bfcaa1ed44b4dcb9124c1d6"; }; # Should be kept in sync with PIN_COQ in CI workflow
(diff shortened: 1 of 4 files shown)


cpitclaudel:
I'm not sure if I follow entirely, but I think your second option is the behavior I expected intuitively.  In Alectryon we had trouble at first because queries we sent were dropped (https://github.com/rocq-prover/vsrocq/issues/1092). We added exponential backoff repeating a query that errored instead of being silently dropped as a way to detect when the server is ready, but I like the idea of queries simply being delayed if the server isn't ready.

PR #1190, opened 2025-12-08
speedup range/position of loc

### LINKED ISSUE #1199: [BUG] VSRocq miscomputes line/column offsets
## **Describe the Bug**

I'm porting Alectryon to vsrocq. VSRocq seems to use two kinds of offsets: UTF-8 offsets in documentState messages, and codepoints / characters (‽) in range information.  The result is that valid documents get VSRocq into an inconsistent state.

## **To Reproduce**

```
Notation "🦔🦔🦔" := 1.
Definition a := 1.
```

### Expected: `rocq top` / Proof general output:

Image

Image

### Actual: `vsrocq output`:

Image

After trying to process one such sentence VSRocq gets into an inconsistent state.  For example, adding a `+1`:

Image

## Notes

The document state is right:

```
Document using sentences_by_id map

[2] [Notation--🦔🦔🦔--:=--1--.] (0 -> 29) (executed)
[3] [Definition--a--:=--1--.] (30 -> 48) (not executed)

Document using sentences_by_end map

[2] [Notation--🦔🦔🦔--:=--1--.] (0 -> 29) (executed)
[3] [Definition--a--:=--1--.] (30 -> 48) (not executed)
```

Note the numbers given in it: `0 -> 29` are UTF-8 offsets.  The first sentence is 20 characters long, but because each 🦔 is 4 bytes in UTF-8, we get a total of 29 bytes (17 ASCII chars + 3*🦔 = 29).

But now compare the messages that VSRocq sends the client:

```
[Trace - 10:50:43 PM] Received notification 'prover/updateHighlights'.
Params: {
    "uri": "file:///home/cpc/git/epfl/systemf/www/test.v",
    "preparedRange": [],
    "processingRange": [],
    "processedRange": [
        {
            "end": {
                "character": 20,
                "line": 0
            },
            "start": {
                "character": 0,
                "line": 0
            }
        }
    ]
}
```

`"character": 20,` is a violation of the LSP spec.  The client (VSCode) sends `"positionEncodings": ["utf-16"]`, which means that it does not support codepoint offsets.  The right position would be `character: 23`, which counts utf-16 code units: 17 code units for the ASCII text, + 3*2 for the hedgehogs.  

The problem is also visible with diagnostics ranges:

Image

```
Notation "🦔🦔🦔" := 1. Check nat.
Definition a := 1.
```

Finally issue can be further confirmed by looking at the `textDocument/didHover` commands that the IDE sends.  Those count utf-16 code units.  Try hovering `nat` below: you'll see that Rocq gives the `Set` hover on the `:= n`, not on `nat`:

```
Notation "🦔🦔🦔" := nat. 
```

(Btw, isn't that range one character too wide?  Should there really be a hover on the `.`? I'll open a separate issue.)

## **Environment**

| Debug Information | Value |
| ----------------- | -------------------------------- |
| Rocq Installation | The Rocq Prover, version 9.2+rc2 compiled with OCaml 5.4.0  |
| Rocq Path         | `/home/cpc/.opam/rocq-9.2/lib/coq ` |
| VsRocq Extension Version   | 2.3.4 |
| VsRocqTop Version | 2.3.4 |
| VsRocqTop Path  | `/home/cpc/.opam/rocq-9.2/bin/vsrocqtop` |
| OS               | x64 linux |
| VSCode Version | 1.109.3 |

gares:


shilangyu:
@gares what is the status here? Is this indeed faster while still being correct? Can I help with finishing this PR?

gares:
Yes, if you want to finish this, please go.
I don't recall if the overall benefit was that huge in practice.

shilangyu:
I added a benchmarking setup to help investigate this change and others in the future. I can split it into a separate PR.

I will next see if this change is indeed noticeable or not.

shilangyu:
First thing I am seeing is that computing document outline/folding ranges becomes 18 times faster on the large example Pff.v file! That is huge, I will look further into it but I believe it will allow us to enable it for all files.

KacperFKorban:
> First thing I am seeing is that computing document outline/folding ranges becomes 18 times faster on the large example Pff.v file! That is huge, I will look further into it but I believe it will allow us to enable it for all files.

That would be amazing!

> I added a benchmarking setup to help investigate this change and others in the future. I can split it into a separate PR.

This would be really useful, I think that we are missing ways to measure performance gains/regressions.

gares:
The bench framework is clearly welcome, I don't mind having it together a pr also using it to demonstrate the benefit of the change.

18x looks amazing, a bit hard to believe to be honest, too good to be true. But well, as long as it is on the plus side, I'm in favor of this change.

gares:
Actually the change is kind of obvious, the only difficulty is getting the integer arithmetic right, and I'm very bad at it.

 So please double check my "halving" code does not do stupid things.

shilangyu:
I will make sure everything is correct before marking the PR ready for review.

I have further improved the performance of pos/loc conversions, leading to outline computations being now 41x faster. If everything turns out to be correct, this will clearly mean we can just enable outline/folding ranges unconditionally. Of course many other things will benefit from these improvements too.

I also think I fixed handling of UTF16 which should address #1199. I am on holidays now so might be a few days before I polish everything. I think it would be nice to include it in the next release, when do you plan to make them @gares?

gares:
I think a release by the end of next week would be nice. I plan to release 9.3.0 by then, but there are some critical issues there, not sure they are all solved yet (I'm on vac too)

shilangyu:
Summary:

- Line search is done using binary search (as implemented by @gares)
- There is an ASCII fast path for conversion between LSP positions and Rocq locations (UTF8 bytes). More often than not lines are just pure ASCII
- UTF16 is handled properly (some code points take two code units to encode, like emojis)
- Added benchmarking
- Fix naming of functions to avoid confusion

Benchmark results before these changes:

```txt
Test Name                              Time
------------------------------------------------------
/entries/big                            763.24 ms/run
/entries/folding_ranges/big             681.28 us/run
/entries/outline/big                    107.12 us/run
/text/edit/big                           64.27 ms/run
/text/parse/big                           2.38  s/run
/text/raw_document/big_create             1.90 ms/run
/text/raw_document/big_loc_of_pos       422.46 us/run
/text/raw_document/big_pos_of_loc         9.88 ms/run
/text/raw_document/big_word_at_loc        2.46 ms/run
/text/raw_document/unicode_loc_of_pos   643.90 us/run
/text/raw_document/unicode_pos_of_loc     2.26 ms/run
/text/raw_document/unicode_word_at_loc    2.43 ms/run
```

And after:

```txt
Test Name                              Time
------------------------------------------------------
/entries/big                             14.83 ms/run
/entries/folding_ranges/big             745.30 us/run
/entries/outline/big                    125.75 us/run
/text/edit/big                           57.48 ms/run
/text/parse/big                           2.40  s/run
/text/raw_document/big_create             2.03 ms/run
/text/raw_document/big_loc_of_pos        28.49 us/run
/text/raw_document/big_pos_of_loc       100.50 us/run
/text/raw_document/big_word_at_loc      106.77 us/run
/text/raw_document/unicode_loc_of_pos   296.34 us/run
/text/raw_document/unicode_pos_of_loc   371.96 us/run
/text/raw_document/unicode_word_at_loc   77.14 us/run
```

Notably:

- Creating document entries (which is used for folding ranges/outline) ~50x faster
- From pos to loc (ascii) ~15x faster
- From loc to pos (ascii) ~98x faster
- Word at location (ascii) ~23x faster
- From pos to loc (unicode) ~2x faster
- From loc to pos (unicode) ~6x faster
- Word at location (unicode) ~31x faster

Ready for review!

shilangyu:
Opened #1309 in case you will find the speedup acceptable enough to enable folding ranges everywhere. @KacperFKorban can you check if the 1.3s you previously reported for the Pff.v file is indeed much faster now?

gares:
I think you got some units wrong in the post above, it seems some stuff got slower before reading your summary.

Amazing!

shilangyu:
Which numbers are you referring to that got slower? There should not be any regressions

gares:
> ```
> /entries/big                            763.24 ms/run
> /entries/folding_ranges/big             681.28 us/run
> /entries/outline/big                    107.12 us/run
> ```
vs
> ```
> /entries/big                             14.83 ms/run (faster)
> /entries/folding_ranges/big             745.30 us/run (???)
> /entries/outline/big                    125.75 us/run (???)
> ```

shilangyu:
I believe this is just noise. None of the changes introduced affect these two benchmarks.

gares:
I'm trying to run the benches and I get:
```
$ dune exec benches/benches.exe
Fatal error: exception File "tests/common.ml", line 188, characters 14-20: Assertion failed
Raised at Tests__Common.handle_d_events in file "tests/common.ml", line 188, characters 14-26
Called from Tests__Common.whole_init_and_parse_test_doc in file "tests/common.ml", line 247, characters 15-46
Called from Tests__Common.init_and_parse_test_doc in file "tests/common.ml", line 250, characters 53-98
Called from Dune__exe__Common.doc_of_str in file "benches/common.ml", line 15, characters 35-95
Called from Dune__exe__Entries.big_outline in file "benches/entries.ml", line 13, characters 12-38
```
this code path is reachable if sel returns None, which happens with a timeout of 0.1s. I guess my disk is slower than yours, or something like that. With 0.2 it then crunches on the bench. Did you have to bump the timeout? how long are the benches supposed to run? Here it has been spinning for 2 minutes already...

gares:
```
Test Name                              Time
------------------------------------------------------
/entries/big                             21.31 ms/run
/entries/folding_ranges/big               1.11 ms/run
/entries/outline/big                    221.75 us/run
/entries/unicode                         14.56 ms/run
/text/edit/big                           64.96 ms/run
/text/parse/big                           4.17  s/run
/text/raw_document/create                 2.12 ms/run
/text/raw_document/loc_of_pos            53.85 us/run
/text/raw_document/pos_of_loc           130.30 us/run
/text/raw_document/unicode_loc_of_pos   254.14 us/run
/text/raw_document/unicode_pos_of_loc   402.21 us/run
/text/raw_document/unicode_word_at_loc  110.31 us/run
/text/raw_document/word_at_loc          138.31 us/run
```

shilangyu:
> Did you have to bump the timeout?

I did not. What I pushed is the state of the repo I had.

> how long are the benches supposed to run?

It runs for about 2min for me. I think it is expected given that it runs them many times. I will see if something is running for longer than it should.

shilangyu:
> Opened #1309 in case you will find the speedup acceptable enough to enable folding ranges everywhere. @KacperFKorban can you check if the 1.3s you previously reported for the Pff.v file is indeed much faster now?

I checked and yes indeed, the speed up is very real.

PR #1194, opened 2026-01-26
Simple MCP server implementation based on vsrocq

KacperFKorban:
This PR adds a base implementation for an MCP server using vsrocq.

The MCP server is implemented as an alternative to the lsp server – passing `-mcp` th `vsrocqtop` starts an mcp server, instead of and lsp server. This way, the user and the LLM don't share the same vsrocqtop session, which should make it easier to delegate tasks.


docs/mcp.md
```diff
@@ -0,0 +1,49 @@
...
+```json
+// .vscode/mcp.json
+{
+	"servers": {
+		"vsrocqmcp_or_some_other_name": {
+			"type": "stdio",
+			"command": "vsrocqtop",
+			"args": [
+				"-mcp",
+				"-coqlib",
```
gares, line 30: I wonder why you need this option. Can't it read the _CoqProject file as vsrocq does?

gares:
thanks for rebasing

KacperFKorban:
Just FYI, I just realized that the new changes are still broken after the rebase, I'm looking into it.

docs/mcp.md
KacperFKorban: Yes, this should definitely be possible in some way. I had plans to separately simplify the initialization phase, but for now I just used the simplest option (not to bundle too many unrelated changes together).
The vscode extension code basically calls vsrocqtop twice, starting with `-where`.
https://github.com/rocq-prover/vsrocq/blob/0ad9d191b7deba12ba6bc240e178d01ab067b7e5/client/src/utilities/toolchain.ts#L103

gares:
I'll merge #1046 and then rebase this one ontop, if it is OK for you

KacperFKorban:
Sounds great! Thanks!

gares:
I'm so sorry I left this on the side for so long. But I did not forget about it.

KacperFKorban:
No problem at all! I just figured that it might be easiest for me to do the rebase. (I might've broken some CI jobs though. I'm looking into that now.)

gares:
I did reorganize the sources so that we have two binaries sharing the document manager, i.e. the apis to check documents, cache states, etc. The two binaries speak via json rpc the LSP and MCP respectively.

I'll look into the tools soon, but I wonder what are your thoughts about the approach of https://github.com/scidonia/rocq-piler by @GavinMendelGleason with whom I recently exchanged at the Iris WS. In particular I recall a discussion about LLM being confused by the column-line addressing approach. Rocq-piler seems to have a nicer approach to this.

KacperFKorban:
Thanks for doing the refactor! I don't have strong opinions about the tools. I definitely thought about implementing more "controlled" tools (in the style rocq-piler has now). But if the line addressing is a problem, one can always disable this one tool (interpretToPoint) and only use interpretToEnd/step*. One thing that I experimentally prefer in the current implementation is no edit tools– different models have very different ways to edit files (patches, sed, scripts. etc.). Also adding more tools adds a cost to the initial context window size.

I think that my general approach is: once we have something that more people can use, we can iterate based on the feedback. And since it is an interface for an LLM we don't have to care about compatibility issues that much.

gares:
The discussion about line-column addressing was either with you @KacperFKorban or with Guillaume about rocq-mcp, not with Gavin. (I feel like my message was not clear)

PR #1228, opened 2026-04-06
Fix string/bool type confusion in client/package.json

cpitclaudel:
This didn't cause issues because `"true"` is truthy in JS and this setting isn't read in OCaml.


gares (approved):

gares:
Thanks

Issue #1225, opened 2026-03-26
[BUG] `> Print` does nothing


mattam82:
**IMPORTANT: Before You Submit This Issue**

- [x] I have searched the [existing VsRocq issues](https://github.com/rocq-prover/vsrocq/issues)
- [x] I have reviewed the [VsRocq FAQ.md](https://github.com/rocq-prover/vsrocq/blob/main/docs/FAQ.md) and my issue is not addressed there.

**Describe the Bug**

I can't see the output of interactive `> Rocq: Print` / `> Rocq: About` commands anywhere and `Rocq: Add Tab` seems to have no effect either to show the query panel where presumably I should see the results.

**To Reproduce**

1. Open a .v file
2. Definition a := 0.
3. In the command pallette: `> Rocq: Print` a
4. Nothing happens

**Environment (paste the result of `Rocq: Troubleshooting: Show Setup`):**

| Debug Information | Value |
| ----------------- | -------------------------------- |
| Rocq Installation | The Rocq Prover, version 9.3+alpha compiled with OCaml 5.4.1  |
| Rocq Path         | `/Volumes/Dev/rocq/rocq-univs/_build/install/default/lib/coq ` |
| VsRocq Extension Version   | 2.3.4 |
| VsRocqTop Version | 2.3.4 |
| VsRocqTop Path  | `_build/install/default/bin/vsrocqtop` |
| OS               | x64 darwin |
| VSCode Version | 1.113.0 |

**Rocq Log Output (if applicable)**

Nothing in `Rocq Log`
In `VsRocq: `
```

[GoalPanel] Rendered in current panel
[GoalPanel] Received proofview notification
[GoalPanel] Rendered in current panel
[GoalPanel] Received proofview notification
[GoalPanel] Rendered in current panel
[GoalPanel] Received proofview notification
[GoalPanel] Rendered in current panel
[GoalPanel] Received proofview notification
[GoalPanel] Rendered in current panel
[GoalPanel] Received proofview notification
[GoalPanel] Rendered in current panel
[DocumentStateViewProvider] Getting document state for uri: file:///Volumes/Dev/rocq/rocq-univs/test-suite/success/telescopes_variances.v
[DocumentStateViewProvider] Firing event
[DocumentStateViewProvider] Current doc uri: file:///Volumes/Dev/rocq/rocq-univs/test-suite/success/telescopes_variances.v
[...]

PR #1229, opened 2026-04-06
Clean up print command in SearchViewProvider (#1225)

cpitclaudel:
I went to read the code when I saw @mattam82's issue.

* client/src/panels/SearchViewProvider.ts (_setWebviewMessageListener): Fix "command" in SearchViewProvider
* client/search-ui/src/App.tsx (handleMessage): Add missing `break` in `printResponse` case.
* language-server/dm/documentManager.ml (print): Wrap body in `try`/`with` to match `about` case.

Introduced in 3c447faf.  Amazingly, `Print` still works, because print responses have the same shape as `About` responses, so handling the former as the latter doesn't cause issues (but then perhaps the code shouldn't be duplicated?).

It might be better to refactor this to get rid of the duplicated code, though.

### DIFF
--- a/client/search-ui/src/App.tsx
+++ b/client/search-ui/src/App.tsx
@@ -74,6 +74,7 @@ const app = () => {
 
             case 'printResponse':
                 handlePrintNotification(msg.data.result);
+                break;
 
             case 'launchedSearch': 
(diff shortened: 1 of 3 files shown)


gares (approved):

gares:
Thanks

PR #1232, opened 2026-04-10
Adapt to rocq-prover/rocq#21912 (add argument to default_flags_of)

yannl35133:
Backwards compatible, but the main PR may need some discussion


gares:
Thanks

yannl35133:
Not needed anymore.

PR #1234, opened 2026-04-11
feat: add -without-project-file flag

### LINKED ISSUE #1233: [BUG] `-boot` flag incorrectly handled
- [x] I have searched the [existing VsRocq issues](https://github.com/rocq-prover/vsrocq/issues)
- [x] I have reviewed the [VsRocq FAQ.md](https://github.com/rocq-prover/vsrocq/blob/main/docs/FAQ.md) and my issue is not addressed there.

**Describe the Bug**

The `-boot` flag which might be present in one's `_RocqProject` is not compatible with a few other flags such as `-where` or `-coqlib`

https://github.com/rocq-prover/rocq/blob/148f045ca589bf214a28314bb6236fff2e60f740/boot/env.ml#L249

VsRocq currently pushes through **all** `_RocqProject` flags whenever the `vsrocqtop` binary is called, I don't think this is correct.

https://github.com/rocq-prover/vsrocq/blob/8a202c061f4490233e2d84b93dba5c6a1f93d4eb/language-server/vsrocqtop/args.ml#L63

Namely, initially VsRocq calls `vsrocqtop -where` to get the rocq path. `vsrocqtop` will gather all flags from `_RocqProject` (including the incompatible `-boot` flag) causing an error.

https://github.com/rocq-prover/vsrocq/blob/0ad9d191b7deba12ba6bc240e178d01ab067b7e5/client/src/utilities/toolchain.ts#L103

**To Reproduce**

I created a small reproduction repo with instructions here: https://github.com/shilangyu/dune-rocq-bug-repro

After cloning, do the following

1. Install dependencies

```sh
opam install --deps-only .
```

2. Build project

```sh
dune build
```

3. Open VSCode with the VSRocq extension.

4. Open the `S.v` file and try to evaluate it with VSRocq. 

**Expected vs. Actual Behavior**

You will the following error:

Image

Instead, the `-where` should succeed.

**Solution**

For setup `vsrocqtopWhere()` (and maybe also `rocqVersion()`?) flags from `_RocqProject` should not be loaded, as I believe they don't affect the outcome.

shilangyu:
This will allow using `vsrocq` in a more controlled way when we do not want `_RocqProject` to affect the outcome.

fixes #1233

### DIFF
--- a/client/src/utilities/toolchain.ts
+++ b/client/src/utilities/toolchain.ts
@@ -103,7 +103,7 @@ export default class VsRocqToolchainManager implements Disposable {
     private vsrocqtopWhere() : Promise<void> {
         
         const config = workspace.getConfiguration('vsrocq').get('args') as string[];
-        const options = ["-where"].concat(config);
+        const options = ["-without-project-file", "-where"].concat(config);
         const cmd = [this._vsrocqtopPath].concat(options).join(' ');
 
         return new Promise((resolve, reject: ((reason: ToolchainError) => void)) => {
--- a/language-server/vsrocqtop/args.ml
+++ b/language-server/vsrocqtop/args.ml
@@ -17,6 +17,7 @@ let Dm.Types.Log log = Dm.Log.mk_log "args"
 let rec skip_xd acc = function
 | [] -> (), List.rev acc
 | "-vsrocq-d" :: _ :: rest -> skip_xd acc rest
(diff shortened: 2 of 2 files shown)


gares (approved):

gares:
Thanks!

shilangyu:
Thank you for such a quick merge!

What is the release schedule of vsrocq? There are a few changes that were merged I am interested to start using (including this one hehe)

PR #1235, opened 2026-04-12
fix: ignore uneeded files from VSIX

### LINKED ISSUE #1147: [BUG] The compiled extension file is large (63.4 Mb)
**Describe the Bug**

The compiled extension file is large because it contains all development dependencies of `pp-display`.

`.vscodeignore` should include rules for `pp-display`.

**To Reproduce**

Download the extension `vsix` file from the Marketplace and check its size.

shilangyu:
Reduces the extension size from 56MB to 700KB.

fixes #1147

### DIFF
--- a/client/.vscodeignore
+++ b/client/.vscodeignore
@@ -11,12 +11,13 @@ vsc-extension-quickstart.md
 **/.eslintrc.json
 **/*.map
 **/*.ts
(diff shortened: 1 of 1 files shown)


gares (approved):

gares:
Thanks

PR #1238, opened 2026-04-14
release 2.4.1

gares:


client/src/utilities/versioning.ts
```diff
@@ -55,7 +55,8 @@ const versionRequirements : VersionReq = {
     '2.3.2': '2.3.0',
     '2.3.3': '2.3.3',
     '2.3.4': '2.3.3',
-    '2.4.0': '2.3.3'
+    '2.4.0': '2.3.3',
+    '2.4.1': '2.3.3'
```
shilangyu, line 59: Is this correct? I think [this PR](https://github.com/rocq-prover/vsrocq/pull/1234) causes the newest extension to not work with a server that does not handle the flag.

gares: You are right, I checked the protocol but not the command line

PR #1240, opened 2026-04-16
2.4.3 with resilient init

gares:
CC @shilangyu

### DIFF
--- a/client/src/utilities/toolchain.ts
+++ b/client/src/utilities/toolchain.ts
@@ -110,48 +110,39 @@ export default class VsRocqToolchainManager implements Disposable {
         const pre240cmd = [this._vsrocqtopPath].concat(pre240options).join(' ');
 
         return new Promise((resolve, reject: ((reason: ToolchainError) => void)) => {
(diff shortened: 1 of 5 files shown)


client/src/utilities/toolchain.ts
```diff
@@ -105,32 +105,44 @@ export default class VsRocqToolchainManager implements Disposable {
...
+
+            exec(cmd, {cwd: workspace.rootPath}, (error, stdout, _stderr) => {
                 if(error) {
-                    reject({
-                        status: ToolChainErrorCode.launchError, 
-                        message: `${this._vsrocqtopPath} crashed with the following message: ${stderr}
-                        This could be due to a bad Rocq installation or an incompatible Rocq version.`
+                    exec(pre240cmd, {cwd: workspace.rootPath}, (error, stdout, stderr) => {
+                        if(error) {
+                            onError(stderr);
```
shilangyu: I think we should concatenate both stderrs. Otherwise debugging why the first exec failed will be difficult as it will always be shadowed by the `-boot` error from the second exec.

gares: thanks

shilangyu (approved):

PR #1241, opened 2026-04-17
chore: modernize client codebase

shilangyu:
See individual commits for the introduced changes. In general, the goal is to not complicate the setup (but rather simplify it) while modernizing the tools and dependencies.

One thing I was not able to do myself is update the nix flakes. I know nothing about Nix to competently deal with it. Help would be appreciated.

Finally, I would like to run prettier since it has been setup in the codebase, but none of the files actually follow the formatting. This will generate a very large diff, so I would prefer to do it in a separate PR.


shilangyu:
~I broke testing, I am investigating it now~ fixed

gares (approved):

shilangyu:
What should we do about the nix files that reference yarn? I am not confident enough to modify them.

gares:
This PR introduced a regression. If I open a file and execute to a point, in the middle of a proof, then the server sends a prover/proofView with contents, but the goal-view-ui app displays "not in proof mode". Re executing to that very point displays the goals. You can also reproduce it by simply closing the goal windows, and executing to a point. It seems the first "update" is lost.

shilangyu:
> This PR introduced a regression. If I open a file and execute to a point, in the middle of a proof, then the server sends a prover/proofView with contents, but the goal-view-ui app displays "not in proof mode". Re executing to that very point displays the goals. You can also reproduce it by simply closing the goal windows, and executing to a point. It seems the first "update" is lost.

I will have a look at it tonight

PR #1246, opened 2026-05-10
Restore hypothesis order in proofView

cpitclaudel:
Follow up to c2052b19, which flipped the order of hypotheses in `prover/proofView`.
Tagging @SkySkimmer for review

### DIFF
--- a/language-server/protocol/ppProofState.ml
+++ b/language-server/protocol/ppProofState.ml
@@ -85,7 +85,7 @@ let mk_goal env sigma g =
   {
     id;
     name;
-    hypotheses = List.rev hyps;
+    hypotheses = hyps;
     goal = Pp.string_of_ppcmds ccl;
   }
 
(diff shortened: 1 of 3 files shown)


language-server/protocol/ppProofState.ml
SkySkimmer: could remove this rev instead

SkySkimmer: (that's what I did for the coqidetop patch, but then I forgot I had done it when making the vsrocq patch but did remember there was a rev somewhere so blindly added it expecting tests to say something if it was wrong)

SkySkimmer (commented):
since there are 2 functions changed shouldn't there be 2 tests?

cpitclaudel:
> since there are 2 functions changed shouldn't there be 2 tests?

Sure.  Fixed.

gares:
Thanks

gares (approved):

PR #1249, opened 2026-05-13
chore: format with prettier

shilangyu:
As mentioned in #1241, the client had prettier configured and it runs if you format files in vscode, but for some strange reason none of the files were _actually_ formatted. This PR runs prettier with `npm run format`. The benefit (other than of course a consistent code style) is that new contributors won't generate large diffs that are unrelated to their changes just because the formatter ran for them when saving a file.


gares:
please rebase

PR #1248, opened 2026-05-13
parse Rocq version 9.2 correctly (fix #1243)

gares:

PR #1250, opened 2026-05-16
Fix version parsing

shilangyu:
Fix problem from #1248


SkySkimmer:
9.2 is in CI, do we not test enough?

shilangyu:
> 9.2 is in CI, do we not test enough?

Is it though?

https://github.com/rocq-prover/vsrocq/blob/f9292274d9295e3ea6d4de6e8a33ff3449737531/.github/workflows/ci.yml#L31

I can add it.

SkySkimmer:
https://github.com/rocq-prover/vsrocq/blob/f9292274d9295e3ea6d4de6e8a33ff3449737531/.github/workflows/ci.yml#L92 ?

shilangyu:
> https://github.com/rocq-prover/vsrocq/blob/f9292274d9295e3ea6d4de6e8a33ff3449737531/.github/workflows/ci.yml#L92
> ?

This job runs only on ubuntu, not macos which is where the problem lies.

gares (approved):

gares:
Please add an osx job

shilangyu:
Added macos and undid the fix to see if the CI catches the problem. I will remove this undo commit afterwards.

shilangyu:
It indeed failed: https://github.com/rocq-prover/vsrocq/actions/runs/25994372604/job/76406159792?pr=1250#step:4:219

shilangyu:
Seems like there are more failures regarding version parsing. https://github.com/rocq-prover/vsrocq/actions/runs/25998571776/job/76417378826?pr=1250#logs

I am not familiar with what is expected and what is a bug. Can we just adjust the regex to be unix compatible rather then potentially using GNU quirks?

language-server/Makefile
```diff
@@ -27,8 +27,8 @@ fatalwarnings: $(DUNE_IN_FILES)
 	@rm -f $@
 	@echo "; generated by make, do not edit\n" > $@
 	version=$$(if command -v coqc >/dev/null; then coqc -print-version; else rocq -print-version; fi); \
-	if [ "$$version" = "9.2" ]; then version="9.2.0"; fi; \
 	version=$${version% *}; \
+	if [ "$$version" = "9.2" ]; then version="9.2.0"; fi; \
 	if echo $${version} | grep -q '^[0-9]\+\.[0-9]\+\.[0-9]\+$$'; then \
 		major_version=$${version%%.*}; version=$${version%.*};\
 	elif echo $${version} | grep -q '^[0-9]\+\.[0-9]\+\(+.*\|\)$$'; then \
```
SkySkimmer:
```suggestion
	elif echo $${version} | grep -q '^[0-9]\+\.[0-9]\++.*$$'; then \
```
the comment above says "Coq/Rocq version is major.minor, then either .patch or +string" so there is no need to try to match major.minor (now that we have the special case for 9.2)

language-server/Makefile
```diff
@@ -27,11 +27,12 @@ fatalwarnings: $(DUNE_IN_FILES)
...
 	version=$$(if command -v coqc >/dev/null; then coqc -print-version; else rocq -print-version; fi); \
-	if [ "$$version" = "9.2" ]; then version="9.2.0"; fi; \
 	version=$${version% *}; \
+	# exceptionally, the version string of Rocq 9.2 is missing the patch version
+	if [ "$$version" = "9.2" ]; then version="9.2.0"; fi; \
 	if echo $${version} | grep -q '^[0-9]\+\.[0-9]\+\.[0-9]\+$$'; then \
 		major_version=$${version%%.*}; version=$${version%.*};\
-	elif echo $${version} | grep -q '^[0-9]\+\.[0-9]\+\(+.*\|\)$$'; then \
+	elif echo $${version} | grep -q '^[0-9]\+\.[0-9]\++.*$$'; then \
 		major_version=$${version%.*}; \
```
shilangyu, line 36: Actually, why don't we also set the `version` itself here too? Otherwise this will produce for example `9.3+alpha` which is not handled by the conditional compilation in OCaml.

shilangyu: @SkySkimmer can I get your input here? Am I missing something?

SkySkimmer: The conditional compilation is always written like `if coq = old1 || coq = old2 then old_code else new_code`, so the current master version does not need to be recognized.
We probably should be removing the trailing `+` to recognize previous rc versions, but I guess nobody uses them so nobody has complained so far.

shilangyu: Fair enough. But I don't see a downside in us treating `9.3+alpha` as `9.3` for instance. It would still hit the else branch but it would allow early preparations for a new rocq release as well as supporting old previous rc version for free.

PR #1253, opened 2026-05-19
fix: synchronize first UI render

shilangyu:
@gares this fixes the issue with the proof view not showing on first try.

vsrocq is operating under the assumption that once the HTML/JS was handed over to vscode, it will be synchronously rendered allowing us to immediately send a message to the webview by assuming that the UI has initialized all of the listeners. In react 18, this synchronous behavior has been retired. So the issue here was that react did not initialize listeners to receive the proof goals data.

In general the best fix would be to not assume this, as this is anyway a very strong assumption that in general does not hold true. But for now I just force a synchronous render at the very beginning just to make sure the listeners are initialized.

### DIFF
--- a/client/goal-view-ui/src/index.tsx
+++ b/client/goal-view-ui/src/index.tsx
@@ -1,12 +1,17 @@
 import React from "react";
-import App from "./App";
+import { flushSync } from "react-dom";
 import { createRoot } from "react-dom/client";
+import App from "./App";
 
 const container = document.getElementById("root")!;
 const root = createRoot(container);
 
-root.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-);
+// we make sure that the first render is synchronous, allowing us to immediately
+// send events to the webview without worrying about the component not being ready yet.
+flushSync(() => {
+  root.render(
+    <React.StrictMode>
+      <App />
+    </React.StrictMode>,
+  );
+})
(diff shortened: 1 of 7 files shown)


gares:
Cc @rtetley

gares:
Thanks!

rtetley:
Nice ! Thanks a lot ! Maybe a simple solution is to have the webview send a handshake to the panel provider (relevant function here: https://github.com/rocq-prover/vsrocq/blob/a8474a0e0f49efbabf169b1f2b6be051095d788b/client/src/panels/GoalPanel.ts#L248)
And until the handshake has been received, queue the messages and dequeue everything once everything is properly initialised ?

shilangyu:
Yes that would be ideal. I can implement that if you want?

rtetley:
That would be awesome ! Thanks !

client/goal-view-ui/src/index.tsx
shilangyu, line 1: Sorry for the diff. It is the autoformatter. I think #1249 would resolve these issues.

shilangyu (commented):
Should be good now!

rtetley:
All good on my end ! Thanks ! @gares do you know why it seems to be failing on macOS for rocq >= 9.0 ?

gares:
I pushed a blind fix for CI, I really don't get why it fails

PR #1256, opened 2026-06-04
feat: textDocument/foldingRange

### LINKED ISSUE #137: Allow folding of Proofs and Sections

### LINKED ISSUE #799: Add support for folding ranges
Fix the folding ranges by letting the server send the ranges contained in ```Proof. [...] Qed.``` Or ```Proof. [...] Admit.```.
See: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange

KacperFKorban:
closes #799
closes #137


.gitignore
```diff
@@ -10,3 +10,5 @@ language-server/tests/vsrocq_config.mlh
 language-server/*.install
 .DS_Store
 result
+*.local*
+_opam
```
shilangyu: Where is a top level _opam getting created?

language-server/vsrocqtop/lspManager.ml
```diff
@@ -432,8 +434,20 @@ let textDocumentCompletion id params =
...
     return_completion ~isIncomplete:false ~items, []
 
+let documentFoldingRange id params =
+  let Lsp.Types.FoldingRangeParams.{ textDocument = { uri } } = params in
+  match Hashtbl.find_opt states (DocumentUri.to_path uri) with
+  | None -> log (fun () -> "[documentFoldingRange] ignoring event on non existent document"); Error({message="Document does not exist"; code=None})
+  | Some { st } ->
+    log (fun () -> "[documentFoldingRange] getting folding ranges");
+    if Dm.DocumentManager.is_parsing st then
+      Error {code=(Some Jsonrpc.Response.Error.Code.ServerCancelled); message="Parsing not finished"}
```
shilangyu: How will this work if we really are in the middle of parsing? The client won't ask us again for the folding ranges once we are done parsing, no?

shilangyu (approved):
💌

.gitignore
gares: It is not unseen in ocaml projects (a bit like node_modules), but maybe language-server/_opam should also be listed. The dev creates it / links it with opam.

language-server/dm/documentEntries.ml
```diff
@@ -0,0 +1,485 @@
...
+        entries_of_whole_constr raw ty @ entries_of_constr raw ty
+      ) constructors
+    | Vernacexpr.RecordDecl (_, fields, _) ->
+      List.concat_map (fun (decl, _attrs) -> entries_of_local_decl raw decl) fields
+  in
+  param_entries @ extra_param_entries @ rtype_entries @ ctor_entries
+
+(** Extracts sub-sentence Gallina folds from vernacular AST nodes that contain
+    located terms. *)
+let entries_of_vernac_ast (document: Document.document) (sentence: Document.sentence) (ast: Synterp.vernac_control_entry) : entry list =
```
gares, line 428: Can you have a look at the outline? (it is roughly a bunch of ranges with a label)

There is quite some code that is common, in spirit at least.

https://github.com/rocq-prover/vsrocq/blob/90a46b4e4269e6c051b7225e6ee998e92dc6dd14/language-server/dm/document.ml#L191-L252

I wonder if it makes sense to have the outline be a projection of this folding info...

.gitignore
shilangyu: > maybe language-server/_opam should also be listed

That is why I asked about the top level `_opam`. Because `language-server/_opam` is already in the gitignore.

language-server/dm/documentEntries.ml
KacperFKorban: I designed the folding code with that in mind, I think that it should be fairly easy to reuse the new folding logic for the outline. I use a more "structured" type as an intermediate datatype that should also preserve the nesting information needed for the outline.
I think that we should be able to slightly extend this datatype and be able to project both the folding ranges and the outline from that. (One thing that would have to be filtered out is the indentation-based "fallback" ranges)

language-server/vsrocqtop/lspManager.ml
KacperFKorban: You're right. I pretty much copy-pasted the code from other requests that require parsing. Maybe we should implement some callback mechanism for those, so that the requests would resume after parsing is done...

.gitignore
KacperFKorban: I use a local switch for the whole `vsrocq` directory, that's why. I can maybe merge them into one pattern.

language-server/dm/documentEntries.ml
KacperFKorban: Do you prefer merging the two features in one PR or is it OK to do a separate one for refactoring the outline?

gares: If it is true that this new code is fast enough and a superset of the other, I think the best way to proceed is to:
1. add to the folding ranges data structure the "labels" used by the overview
2. write the projections
3. remove the old overview code and call the new one

Then we merge 1+2 in a PR. Finally we replace the overview code with 3.
If 3 turns out be not be ok, we can just revert one PR and still keep the folding.
But I think it is easier to develop 3 atop the other changes, and just (not) merge the last commit.

language-server/vsrocqtop/lspManager.ml
gares: I don't know about folding ranges, but for the overview we do return the message "not ready yet" but vscode never asks again. It seems a bug in vscode (or better, it does not implement LSP as designed).

KacperFKorban: Hmmm, IIUC the [`ServerCancelled`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#errorCodes) error is only supported for semantic tokens and diagnostics.

shilangyu: Should parsing be an asynchronous task then? It seems reasonable to me for the LSP to wait until it is done with parsing before responding to a request.

KacperFKorban: That seems reasonable. Still, I would implement that in a new PR, and allow for returning ServerCancelled here for now.

gares:
apparently `CRecord` changed signature, here the doc of ppx_optcomp:

https://github.com/janestreet/ppx_optcomp#conditionals

there is a way to "version" a pattern match branch `| CRecord x [...] -> | CRecord(x,y) [...] ->`

language-server/dm/utilities.ml
```diff
@@ -66,3 +66,134 @@ let vernacstate_synterp_parsing x = x.Vernacstate.synterp.Vernacstate.Synterp.pa
...
+  | Constrexpr.CDelimiters (_scope, e) -> fold_constr f acc e
+  | _ -> acc
+[%%else]
+let fold_delimiters fold_constr f acc e =
+  match e.CAst.v with
+  | Constrexpr.CDelimiters (_depth, _scope, e) -> fold_constr f acc e
+  | _ -> acc
+[%%endif]
+
+let rec fold_constr (f : 'acc -> Constrexpr.constr_expr -> 'acc) (acc : 'acc) (e : Constrexpr.constr_expr) : 'acc =
```
SkySkimmer: Is this very different from Constrexpr_ops.fold_constr_with_binders?

KacperFKorban: Oh, it might be just the function I wanted. Just couldn't find the right API. I'll look into it tomorrow.

KacperFKorban: Yes, this was the exact function that I wanted. I should've looked a bit more before I re-implemented it. Thanks!

.gitignore
```diff
@@ -1,3 +1,13 @@
+language-server/_build
+_opam
+language-server/dm/dune
+language-server/language/dune
+language-server/protocol/dune
+language-server/vsrocqtop/dune
+language-server/tests/dune
+language-server/dm/vsrocq_config.mlh
+language-server/tests/vsrocq_config.mlh
+language-server/*.install
```
shilangyu: Leftover from a merge conflict

language-server/dm/utilities.ml
```diff
@@ -66,3 +66,20 @@ let vernacstate_synterp_parsing x = x.Vernacstate.synterp.Vernacstate.Synterp.pa
 [%%else]
 let vernacstate_synterp_parsing x = Vernacstate.(Synterp.parsing x.synterp)
 [%%endif]
+
+let option_to_list = function
```
shilangyu, line 70: I think that already exists https://ocaml.org/manual/5.5/api/Option.html#VALto_list

language-server/vsrocqtop/lspManager.ml
```diff
@@ -440,8 +442,20 @@ let textDocumentCompletion id params =
...
+    else
+      let folding_ranges = Dm.DocumentManager.get_folding_ranges st in
+      Ok(Some folding_ranges)
```
shilangyu, line 461: I don't know what are the agreed upon API boundaries, but should DM be returning LSP types?

language-server/tests/callgraph/static_callgraph_whitelist.txt
```diff
@@ -28,6 +28,8 @@ Names\.Id\.
 Loc\.
 CAst\.make
 Future\.create_delegate
+Constrexpr_ops\.fold_constr_expr_with_binders
+Constrexpr_ops\.fold_constr_expr_with_binders\..*
```
shilangyu: Shouldn't these be merged into one?

language-server/dm/documentEntries.ml
```diff
@@ -0,0 +1,666 @@
...
+
+let names_or_default names =
+  match names with
+  | [] -> ["default"]
+  | _ -> List.map Names.Id.to_string names
+
+let declaration_detail (raw: RawDocument.t) start stop : string =
+  let source = RawDocument.string_in_range raw start stop in
+  let max_characters = 100 in
+  let ellipsis = "..." in
```
shilangyu: An unicode ellipses would be better I think

language-server/dm/documentEntries.ml
```diff
@@ -0,0 +1,666 @@
...
+      name = outline_info.outline_name;
+      detail = outline_info.detail;
+      kind;
+      range = entry.range;
+      selectionRange = outline_info.selection_range;
+      children;
+      deprecated = None;
+      tags = None;
+    }]
+  | _ -> children
```
shilangyu, line 658: I feel like these should happen outside of `dm/`

shilangyu: Actually no, I see it is imported in other `dm/` files. So I don't understand the split too much

language-server/dm/documentManager.ml
```diff
@@ -214,64 +215,24 @@ let get_document_proofs st =
...
+    | Some entries -> entries
+    | None ->
+      let entries = DocumentEntries.entries st.document in
+      st.folding_entries_cache := Some entries;
+      entries
 
 let get_document_symbols st =
-  let outline = List.rev @@ Document.outline st.document in
-  get_document_symbols outline [] []
+  DocumentEntries.document_symbols (entries_for_request st)
```
shilangyu, line 208: This means we are replacing the previous outline implementation? Does it mean we can remove the previous code that did it?

shilangyu (commented):
A few questions/comments

language-server/dm/utilities.ml
KacperFKorban: `Option.to_list` doesn't work for me

language-server/vsrocqtop/lspManager.ml
KacperFKorban: Hmmm, doesn't it do so already? I don't fully understand the split either. My intuition was that the conversions to LSP are usually done in documentManager

language-server/dm/documentManager.ml
KacperFKorban: Looks like the was another feature depending on it `get_document_proofs`, now I have removed it entirely.

language-server/dm/documentEntries.ml
KacperFKorban: I'm not sure what the correct boundary is. I didn't move the implementation for now.

shilangyu (approved):
Amazing work :) Looks great!

(I did not test if the new documentProofs works)

gares:
Looks great, my only concern is how fast it is on large files.

We already have perf issues, and some of them have some hope to be solved by not reparsing everything (I never finished the PRs, but they are there). If one does not reparse, he has to adjust the existing data by typically shifting.

Can you please test it on a large file? I usually use pfff.v 
[Pff.v.gz](https://github.com/user-attachments/files/30699410/Pff.v.gz)
The point is just to measure "how much slower" it is.
And if it is actually slower, would the strategy above (not recomputing but shifting) applicable?

gares:
Or if you have better mitigation strategies, just share...

KacperFKorban:
@gares The initial request takes ~1.3s for this file, so it is fairly significant. Afterwards, if we don't need to recompute, then it's 3ms, because it's all cached.
Though on these big files, the behavior of returning an error on incomplete parsing seems very visible – very often the client fallback is used, since the parsing takes so long. (Fixing this might be one of the next things that I will look into)

I see a few immediate options here:
1. Wait with merging this feature until we have a better handling for re-parsing.
2. Merge it, but add a config to disable it.
3. Same as `2.`, but make it opt-in, not out.
4. Come up with some heuristic that would only use the semantic folding ranges for files that are < ~3000 loc.

I would prefer to avoid `1.`. Maybe some combination of `2.-4.` might work?

gares:
I would prefer 4 and have the cutoff configurable

shilangyu:
How does the 1.3s compare to the previous implementation (computing only the outline)? And that 1.3s, is just to compute document entries, or does it include parsing done by Rocq (which we cannot avoid anyway, right?).

Do we know how often the LSP client asks for folding ranges/outline? Is it after every edit with some debounce? I am asking to know if this 1.3s happens after every edit or not. Since vsrocq server has some event system, does it somehow allow for the LSP to handle other requests while the document entries are computed or not really and everything is blocked?

Personally, I don't like the 4th option. I find it very strange to disable features due to file size.

gares:
I think there is a request after each edit.
From the previous message, I assumed 1.3s in addition to parsing (on my slow laptop parsing is like 3s).

gares:
file size cap = 0 means disabled, 999999 means enabled, anything in between is a default that does not make the thing unusable unless you tweak the conf file. Maybe the problem is then discoverability (how do you know the thing is cut off and not broken...)

gares:
CI green and PR mergeable.

@shilangyu do you have a better proposal for the cutoff? Maybe you would prefer a time based cutoff? I do not have a strong opinion, I just want to avoid regressions (in performance) and still merge this PR before the next release

gares (approved):

shilangyu:
> @shilangyu do you have a better proposal for the cutoff? Maybe you would prefer a time based cutoff? I do not have a strong opinion, I just want to avoid regressions (in performance) and still merge this PR before the next release

I think just merging this as is makes most sense. Personally, I am not happy with having a cut off but I understand getting something better will require significantly more work.


One thing I would comment on is the choice of "character" as a unit and the default of 3000: I checked and almost every file in my project easily surpasses 3000 characters (it is about 90LOC). Plus characters are a rather abstract unit for the user. I feel like LOC or sentence count would be easier to understand as a configuration. I think sentence count is the best predictor for computation time? Though still LOC seems the most intuitive to me.

shilangyu:
Also, I just realized the config says "loc" which may be confused with LOC when it fact means "location". Loc is an internal name, I don't think the user will know what that is (nor should they).

shilangyu:
It should also be noted that a cutoff means LSP requests fail producing a vscode popup:
image

gares:
Hum, this has to be fixed. Also I thought it was Lines, not characters. 3k Lines is reasonable, 3k chars is clearly not.

KacperFKorban:
The characters vs loc is a mistake. Thanks for spotting that. I'll fix that tomorrow.

About returning an error for the limit, I don't have a good idea for a better solution yet. I'll try to think of something.

gares:
The overview, when run late, would simply not be there instead of giving an error. No idea how this was achieved

shilangyu:
Returning `null` as the result should yield the desired behavior

gares:
I tested it a little and it seems to work as expected.
One thing, I noticed that, with the old extension (client code), the overview does not work, while it works with the new one.
This coupling has to be reflected in the versioning.ts file.
I don't think this is a problem, after all this will be a big release, I just want to be sure this was intended.

shilangyu:
> with the old extension (client code), the overview does not work

I think it is a bit surprising, this changeset only affects the LSP so it should just work out of the box even with the old version.

PR #1261, opened 2026-06-12
fix: also enqueue block_events in execution_finished

### LINKED ISSUE #1252: [FEAT] Jump to line of error during an error
**Is your feature request related to a problem? Please describe.**

When defining a term (not in proof mode) that spans across many lines, if there is a type error, stepping will show the error but still jump to the end of the definition, which is usually far from the exact line of error.

**Describe the Solution You'd Like**

It would be nice to automatically focus the editor to the line or error, and bring attention to the user, for example by highlighting.

**Describe Alternatives You've Considered**

I could scroll manually to the error after stepping, edit, step, and scroll again, but it's hard to debug/iterate that way.

KacperFKorban:
closes #1252

### DIFF
--- a/language-server/dm/checkingManager.ml
+++ b/language-server/dm/checkingManager.ml
@@ -499,14 +499,14 @@ let interpret_in_background document st =
 let validate_document document st =
   if !settings.check_mode = Settings.Mode.Continuous then interpret_in_background document st else st, []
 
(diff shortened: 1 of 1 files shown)


gares (approved):
Thanks, such a silly mistake I made here

PR #1264, opened 2026-06-15
fix: wait for UI to send display settings

### LINKED ISSUE #1260: [BUG] possible regression in configuration update
The goal printing depth as configured is not kept into account by vsrocq upon window reload unless the value is changed.

**To Reproduce**

- execute up to a goal
- you see some `...`
- change the printing depth to a high value
- the goal shows less `...`
- F1 reload window
- go to the same goal
- the `...` are back
- change the printing depth to a different value, but still high
- the `...` are gone

@thery could you please check if the steps above correspond to your description?

shilangyu:
Fixes #1260

We must wait for the UI to finish initializing before sending the display configuration to it. Otherwise the message is lost. See #1253 for more explanation.

I additionally:
- Added type safety to many bare strings and payloads which were waiting for a typo
- Made the goal view react to changes in the maxDepth. Before, it required a refresh of the goal view


gares:
Sorry, I must have merged another PR. please fix the conflict.

shilangyu:
Should be good now

gares:
maybe @rtetley can give a second look at the ts code.
Otherwise I will test and merge

shilangyu:
@KacperFKorban can you test this branch and let me know if it seems fine to you? If so, then we can merge it. On my side this seemed to fix the bug.

client/src/package-json.ts
KacperFKorban, line 1: Is the convention to commit generated TS types files? (Just making sure that it's not by accident)

KacperFKorban (approved):
I tested it locally and it seems to work. The code looks good.

client/src/package-json.ts
shilangyu: Good question. We could make it part of the build and gitignore it. Then the CI would see if something is outdated. I will do that.

shilangyu: ^Did that

shilangyu:
CI failure seems unrelated. If this looks good to you @KacperFKorban then feel free to merge.

PR #1266, opened 2026-06-17
Include css in the pp-display package

### LINKED ISSUE #1265: [BUG] Regression in message display
In master we have a regression wrt 2.4.3 when it comes to the display of messages in goal window.
I suspect that messages are formatted server side (unlike the goal), inserting \n at breaking points, and somehow that character is dropped in master.

Image
Image

shilangyu:
fixes #1265 

My bad, I too eagerly removed it during the refactor thinking it is no longer needed.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -268,6 +268,43 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@emnapi/core": {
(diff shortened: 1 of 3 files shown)


gares (approved):

gares:
Thanks

PR #1267, opened 2026-06-24
fix: enable the unicode flag in the wordPattern regex

KacperFKorban:
The `wordPattern` regex that is responsible for filtering out the tokens that should be viable for `text` completions uses `\p` which requires the `u` (unicode) flag. Apparently this isn't enabled in vscode? (Though the rust regex-regex documentation says that it should be on by default)

This caused text based completions to often suggest completions like `{A`. That is because every part of `\p{L}` was treated as a separate character.

CC regex expert: @shilangyu


client/syntax/rocq.language-configuration.json
```diff
@@ -37,7 +37,7 @@
         }
     ],
     // This allows for VSCode to suggest better intellisense hints (especially for definition names that include a "'" in them)
-    "wordPattern": "(_|[a-zA-Z\\p{L}])([a-zA-Z\\p{L}\\d'_])*",
+    "wordPattern": "(?u:(?:_|[a-zA-Z\\p{L}])(?:[a-zA-Z\\p{L}\\d'_])*)",
```
shilangyu: Doesn't `\p{L}` subsume `[a-zA-Z]`? This should be equivalent (and simpler):

```suggestion
    "wordPattern": "(?u:[_\\p{L}][\\p{L}\\d'_]*)",
```

I am then surprised that vscode's highlight feature (https://github.com/rocq-prover/vsrocq/pull/1262) is not using this to segment the words.

shilangyu (commented):
My regex super senses have examined this PR and produced a review.

Best regards,
Ludwik

client/syntax/rocq.language-configuration.json
KacperFKorban: > Doesn't \p{L} subsume [a-zA-Z]? This should be equivalent (and simpler):

That was my intuition as well, and also my original fix. But my second thought was that maybe explicitly adding the `a-zA-Z` would make the happy case faster... That's why I'm asking the expert ^^

shilangyu: > would make the happy case faster

No, both generate exactly the same IR representation

```sh
$ diff <(regex-cli debug hir '(?u:\p{L})') <(regex-cli debug hir '(?u:[a-zA-Z\p{L}])')
# nothing
```

gares:
Can we get a few tests?

KacperFKorban:
> Can we get a few tests?

I'm not entirely sure what test we could write for this... In the end the fix is most likely caused by a configuration issue/decision in vscode– disabling unicode flag in wordPattern regexes.

KacperFKorban:
I tried adding integration tests for this, but they seemed flaky. This fix is also mainly based on a configuration parameters/dependencies in vscode itself, so it will be really hard to add a test for it.

@shilangyu Could you try testing it locally to see if it works?
This is a snippet that fails for me without a fix:
```rocq
Definition foo {A: Type} (x: A) : A := x.

```
If in the 2nd line, you ask for completions, the word-based completions should incorrectly suggest `{A` without the fix.

shilangyu:
Works properly when tested locally 👍

client/syntax/rocq.language-configuration.json
shilangyu: These two regexes seem to be too defining a regex for identifiers, maybe we should update them too?

shilangyu (approved):

KacperFKorban:
@shilangyu could you take a quick look at the last changes. I applied your comment plus added some additional configuration to the rocq language config that was only present in the coq one (which is AFAIK unused)

client/syntax/rocq.language-configuration.json
```diff
@@ -34,20 +34,64 @@
...
             "close": "|"
+        },
+        {
+            "open": "'",
+            "close": "'",
+            "notIn": [
+                "string",
+                "comment"
+            ]
+        },
```
shilangyu: Do we want that? I don't think single quote strings exist in Rocq (and it might give a false impression that it exists)

client/syntax/rocq.language-configuration.json
```diff
@@ -34,20 +34,64 @@
...
+            "close": ")"
+        },
+        {
+            "open": "'",
+            "close": "'",
+            "notIn": [
+                "string",
+                "comment"
+            ]
+        },
```
shilangyu: Then also here not needed

shilangyu (approved):
Except small comment LGTM

PR #1270, opened 2026-07-06
fix: don't delegate proofs that don't guarantee opacity

### LINKED ISSUE #703: Crash During Fast Edits With Proof Delegation Enabled
If I have the `Proof:Delegation` setting set to `Delegate proofs` and  I quickly edit the `H_edit_me` name (e.g. by holding down a key) in the following file...
```Coq
Require Import Lia.

Derive f SuchThat True As H_edit_me.

Lemma lemm : True.
Proof.
  trivial.
Qed.

Fixpoint pow (a b : nat) : nat :=
  match b with
  | 0 => 1
  | S b' => a * pow a b'
  end.

Lemma computation : (pow 3 21) < 900.
Proof.
  simpl. lia.
Qed.

Lemma computation2 : (pow 2 27) < 900.
Proof.
  simpl. lia.
Qed.

Lemma computation3 : (pow 4 16) < 900.
Proof.
  simpl. lia.
Qed. 
```
, I get an output from the Coq Language Server like
```
[Info  - 5:16:08 PM] Connection to server got closed. Server will restart.
[Info  - 5:16:09 PM] Connection to server got closed. Server will restart.
[Info  - 5:16:11 PM] Connection to server got closed. Server will restart.
[Info  - 5:16:11 PM] Connection to server got closed. Server will restart.
[Error - 5:16:12 PM] Connection to server got closed. Server will not be restarted.
```
and VSCode tells me
image

Then I have to restart VSCode to get the language server running again.

This doesn't happen with `Proof:Delegation` set to `No delegation`.

KacperFKorban:
possibly closes #703

In delegation mode, we look at the VtQed type to determine if the proof has side effects, but some proof "openers" also don't guarantee opacity. We added a check to not delegate the proof if it doesn't guarantee opacity.

### DIFF
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -233,9 +233,9 @@ let interp_qed_delayed ~doc_id ~proof_using ~state_id ~st =
   let control = [] (* FIXME *) in
   let opaque = Vernacexpr.Opaque in
   let pending = CAst.make @@ Vernacexpr.Proved (opaque, None) in
-  (*log fun () -> "calling interp_qed_delayed done";*)
+  log (fun () -> "calling interp_qed_delayed done");
   let interp = Vernacinterp.interp_qed_delayed_proof ~proof ~st ~control pending in
-  (*log fun () -> "interp_qed_delayed done";*)
+  log (fun () -> "interp_qed_delayed done");
   let st = { st with interp } in
   st, success st, assign) 
   |> Result.fold ~ok:(fun x -> x) ~error:(fun x -> CErrors.user_err x)
--- a/language-server/dm/executionManager.ml
+++ b/language-server/dm/executionManager.ml
@@ -233,9 +233,9 @@ let interp_qed_delayed ~doc_id ~proof_using ~state_id ~st =
   let control = [] (* FIXME *) in
   let opaque = Vernacexpr.Opaque in
   let pending = CAst.make @@ Vernacexpr.Proved (opaque, None) in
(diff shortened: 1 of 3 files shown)


KacperFKorban:
The CI failure seems the same as on `main`

gares (approved):
Thanks, that was clearly an oversight

gares:
I let you merge so we test the settings. Please make a merge commit.

SkySkimmer:
I'm seeing related-looking errors in rocq CI https://gitlab.inria.fr/coq/coq/-/jobs/7567716
~~~
File "tests/s_tests.ml", line 53, characters 0-415: schedule: opener without opacity guarantee threw
(runtime-lib/runtime.ml.E "predicate failed"
  ((Value (Error "more errors than expected, extra 2"))
    (Loc tests/common.ml:51:15))).
~~~

KacperFKorban:
@SkySkimmer Thanks for spotting that, I'll look into it right now.

KacperFKorban:
Ah, looks like `rocq-stdlib` isn't installed in this one CI job. I'll add it in a new PR.

SkySkimmer:
not needing stdlib is nice, can't you do the test without it?
Seems to be just for Derive, you can require Derive from Corelib

KacperFKorban:
That works too. I just thought that it might've been an oversight, since coq-stdlib is installed for 8.x in the CI

SkySkimmer:
there's no separate corelib before 9.0

PR #1278, opened 2026-07-06
Do not highlight unicode characters

### LINKED ISSUE #905: Syntac highlighting adds an odd box around lower-case unicode letters
I often use unicode letters in my Coq identifers, to better match what we do on paper. This works fine in Coq, but the syntax highlighting of this extension draws a box around some these characters:

image

This makes them look like some sort of error, and also makes them hard to read.

In a plain text file these letters look just right, so it seems to be something specific to what VsCoq does in `.v` files.

shilangyu:
Fixes #905.

Lean does the same thing: https://github.com/leanprover/vscode-lean4/blob/17d1d086d9cce16f885dde102adb056cad15cb50/vscode-lean4/package.json#L1748

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -311,6 +311,11 @@
         }
       }
     ],
+    "configurationDefaults": {
+      "[rocq]": {
+          "editor.unicodeHighlight.ambiguousCharacters": false
+      }
+    },
     "commands": [
       {
         "command": "extension.rocq.showManual",


Durbatuluk1701:
I personally would oppose this, although maybe I am in the super-minority here. 

My personal opinion is that extensions should do as much as possible to avoid forcing (or maybe not *quite* forcing in this case, but changing users' defaults) configuration settings. 

I also think the default behavior can be quite helpful for distinguishing characters, e.g. 

image

<details>
<summary>Hint</summary>
image

</details>

shilangyu:
In general I agree with you that we should avoid changing user's configs. But with this one I feel that it is quite common to trigger false positives from this setting in languages like Rocq where unicode characters are common.

Though to be fair, the unicode usage seems much more tame compared to Lean codebases. So maybe indeed this is not needed for Rocq?

shilangyu:
@gares @KacperFKorban thoughts? Maybe indeed using crazy unicode is not that common in Rocq and this is setting does more harm than good

KacperFKorban:
I am not opposed to this change. This is also just a default, so can always be overwritten, right?

TheoWinterhalter:
The point of highlighting those characters is to prevent against Pollack inconsistencies is it not? It probably better to leave the default as-is and let users who know what they do (and about the problem) change the default for themselves.

shilangyu:
> can always be overwritten, right?

yes


I am now convinced that the highlighting is not such a problem. I will leave things as is then.

PR #1280, opened 2026-07-08
bump pin rocq

gares:


gares:
I can't fix the nix issue. The problem is that, for rocq master, we need a new dune, >= 3.21.
@rtetley any ideas on how to achieve that?

shilangyu:
I guess this has been now fixed in #1285 ?

PR #1284, opened 2026-07-09
Fix conflicting yojson and ppx_yojson_conv_lib dependencies

VojtechStep:
@gares as I wrote on Zulip this resolves the duplicate dependencies, but then fails to build on master anyways, because it can't find rocq-runtime or something. But I think that needs someone more familiar with dune/ocaml than me to take a look.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,27 +20,27 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1774244481,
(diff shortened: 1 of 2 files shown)


gares:
I think #1285 supersedes this one (and I've no clue what the problem here is)

gares:
Thanks!

PR #1285, opened 2026-07-09
Nix stuff

Durbatuluk1701:
This seems to fix the CI (although I am not sure it is the optimal solution and was patched together after a couple attempts).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -34,6 +34,22 @@
         "type": "github"
       }
     },
+    "nixpkgs-26-05": {
+      "locked": {
+        "lastModified": 1783389287,
+        "narHash": "sha256-0xIy4dVLqq47rA+mRy0hXDfjhQd4E5PoIns/RmB7nR4=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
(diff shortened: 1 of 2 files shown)


Durbatuluk1701:
Related to #1284 and #1280

language-server/dm/executionManager.ml
```diff
@@ -212,7 +212,7 @@ let add_using proof proof_using lemmas =
         using;
       let _, pstate = Declare.Proof.set_used_variables proof ~using in
       pstate
-[%%elif rocq = "8.20" || rocq = "9.0" || rocq = "9.1" || rocq = "9.2"]
+[%%elif rocq = "8.20" || rocq = "9.0" || rocq = "9.1" || rocq = "9.2" || rocq = "9.3" || rocq = "9.3+alpha"]
```
SkySkimmer: there's no way this is correct

SkySkimmer: https://github.com/rocq-prover/rocq/blob/v9.3/vernac/declare.mli#L285 takes an option and doesn't return a pair

Durbatuluk1701: thanks, yeah I had the wrong rocq pin

gares:
Thanks!

PR #1300, opened 2026-08-19
test: wait on published diagnostics instead of sleeping for 10 seconds

TDiazT:
Every test in the client suite waited with a fixed `sleep(10000)` and then read whatever diagnostics happened to be there. This PR changes it to subscribe to diagnostic changes instead.
Specifically, `common.waitForDiagnostics` subscribes to `vscode.languages.onDidChangeDiagnostics`, re-evaluates a test-supplied predicate on every change for the document, and rejects on timeout instead of proceeding with a stale reading.

The wait condition and the assertion must be different predicates. The wait condition is only there to decide when to stop waiting (might not only be time but some specific behavior depending on the fixture), while the assertion is the actual claim we are testing for. If the wait condition is the same as the assertion, then the assertion is redundant.

Waiting on the first publication is not enough for every fixture. `update_view` runs once per executed sentence and publishes the diagnostics of the whole document each time, so a fixture that fails in more than one place is observable in a partial state. `delegate_proof.v` is one: it fails at `apply foobar.` and again at `Qed.`, and a test that resumes on the first of those reads the wrong error. `common.diagnosticOnLine` gives the two tests over that fixture a terminal condition instead, such that a diagnostic on the last sentence means checking reached the end of the document. It still fixes only *where* the server must have spoken, leaving *what* it said to the assertion.

The configuration updates are now awaited. `getConfiguration().update` returns before the setting has reached the server, so an unawaited call leaves the document free to be checked under the previous delegation mode (or under the declared default, `None`, which checks the proof body that `Skip` is meant to drop).

This harness cannot establish the *absence* of diagnostics, which is documented in the helper rather than worked around. An empty array is also the state before the server has published anything, so `d.length === 0` holds from the first instant. Telling "checked, no errors" apart from "not checked yet" needs the progress ranges of `prover/updateHighlights`, which the extension consumes for its decorations and does not expose to the `vscode` API.

Measured on macOS arm64, VS Code 1.134.0. The suite goes from 50s to 0.7s, with 5 tests passing.
Measured on Linux x86_64 (Arch, i5-12600KF), VS Code 1.134.0, Rocq 9.2: The suite goes from 51s to 1.5 s, with 5 tests passing.

### DIFF
--- a/client/src/test/suite/common.ts
+++ b/client/src/test/suite/common.ts
@@ -14,6 +14,94 @@ export async function openTextFile(file: string): Promise<vscode.Uri> {
     return docUri;
 }
 
-export async function sleep(ms: number) {
-    return new Promise((resolve) => setTimeout(resolve, ms));
+/**
+ * Waits until the diagnostics published for `uri` satisfy `predicate`, and
+ * returns them. Rejects after `timeoutMs` instead of continuing with whatever
+ * happens to be there.
+ *
+ * The wait condition and the assertion must be two different statements. The
+ * wait is only a synchronisation point — "the server has said something about
+ * this document" — and the assertion is what the test is actually about
+ * (count, severity, message). A predicate that repeats the assertion makes the
+ * test vacuous: it can only be reached by already being true.
+ *
+ * Known limitation: this cannot establish the *absence* of diagnostics. An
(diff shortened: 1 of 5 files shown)


gares (approved):
Nice

PR #1304, opened 2026-08-24
ci: schedule fewer macOS cells on pull requests

TDiazT:
This is a proposal to cut the macOS matrix on pull requests from 16 cells to 4, one queueing wave instead of four, because GitHub caps concurrent macOS runners at 5 and those cells set the duration of the whole run. Locally tested it went from 26-30 minutes down to 10-12 minutes for the complete workflow (to check how it goes here).

Ubuntu is left untouched so that it runs every combination (Github limits concurrent runners to 20, and we don't reach the limit), and every other event plus any pull request labelled `ci-full` still gets the complete matrix. This requires adding the `ci-label` to the repo if we want to trigger it.

The two routes get different version lists because the older opam macOS cells fail intermittently on arm64 (e.g. the 8.19.0 runs), so opam starts at 8.20.0 while nix keeps the remaining versions that were left out (8.18 and 8.19).
ADR-0001 records the reasoning and the suspected cause, so that restoring the symmetry does not quietly reintroduce the flake. 

The cells are generated by `.github/scripts/ci-matrix.py` because a cross product `os x coq` cannot express per-OS version lists, and running it with `FULL=true` reproduces today's matrix cell for cell. The trigger gains the labeled type, without which applying the label fires no run at all.

### DIFF
--- a/docs/adr/0001-macos-ci-fast-path-is-asymmetric-between-nix-and-opam.md
+++ b/docs/adr/0001-macos-ci-fast-path-is-asymmetric-between-nix-and-opam.md
@@ -1,0 +1,95 @@
+# ADR-0001: The macOS CI fast path is deliberately asymmetric between the nix and opam routes
+
+**Status:** TBD
+**Date:** 2026-08-24
+
+## Context
+
+Most of a CI run is macOS jobs waiting for a free runner.
+For instance, [run 32170095477](https://github.com/rocq-prover/vsrocq/actions/runs/32170095477) for a PR took 26 minutes end to end, and what ended it was `nix-dev-build (macos-latest, 9-1, dev)`, which spent 19.9 of those minutes queued and 6.2 running.
+Its 17 macOS jobs averaged 9.5 minutes queued against 6.2 running, while the 20 ubuntu jobs of the same run averaged 0.5 minutes queued, 3.1 at worst.
+Similarly, [run 32649544820](https://github.com/rocq-prover/vsrocq/actions/runs/32649544820), a push to `main` five days later, has the same shape, 26.7 minutes end to end with 7.7 minutes queued against 5.1 running on macOS and 0.2 queued on ubuntu.
+
+`ci.yml` schedules 16 macOS cells per run, 8 Rocq versions × 2 independent build routes (`nix-dev-build` and `install-opam`), and GitHub caps concurrent macOS jobs at 5 on the Free, Pro and Team plans ([job concurrency limits](https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners)), so those 16 cells execute in 4 waves of about 6 minutes.
+The ubuntu cap is 20 and is never reached.
+
+The two routes do not cover the same ground:
+
+- **nix** builds Rocq from nixpkgs and is the only place where tests actually execute on Darwin (the Electron suite and the LSP suite over stdio).
+- **opam** builds `rocq-core` from source with the system toolchain.
+  On macOS it runs no tests at all, since every test step in `install-opam` is gated on `runner.os == 'Linux'`.
+  Its entire value on Darwin is as a build check of the opam packaging route.
+
+The opam route fails intermittently on older macOS cells.
+What is observed is that the `install-opam` macOS cell for `coq-core` 8.19.0 fails part of the time, with the OCaml compiler aborting during the build, and that the same cell passes on a rerun.
+The nix route has never shown it.
+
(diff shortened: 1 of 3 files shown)


docs/adr/0001-macos-ci-fast-path-is-asymmetric-between-nix-and-opam.md
```diff
@@ -0,0 +1,95 @@
...
+The nix route has never shown it.
+
+The suspected cause is the `build-env` the `coq-core` opam package sets, `OCAMLPARAM = "_,w=-46,warn-error=-a,keywords=5.2"`.
+Compiler-libs binaries parse `OCAMLPARAM` at startup, and `keywords=5.2` means nothing to a compiler older than 5.2, which appears to reach an `assert false` in `utils/warnings.ml`.
+That would make the affected range `coq-core` 8.17.0 through 8.19.1, with 8.19.2 and later unaffected because the variable is added in opam packaging rather than being present in the Rocq source tarball, and it would explain why 8.18.0 sits inside the range yet is not seen failing, the fault being sensitive to the exact argv, environment and paths of each build.
+
+The mechanism has not been confirmed end to end, and only the 8.19.0 cell has been seen failing.
+The decision below rests on the observation that those cells are unreliable on Darwin, and holds whatever the explanation turns out to be.
+If opam's `build-env` is the cause there is no CI-side workaround, since it overrides the ambient value.
```
TDiazT, line 33: I've been trying to debug it with Claude and the closest I got to getting rid of the build error on macOS was with Rocq version 8.19.2, where these flags were removed. 
Maybe you are more familiar with the problem, @gares ?

gares:
I'm not so sure I understand the approach from the description. What does "add a label to the repo mean"?

docs/adr/0001-macos-ci-fast-path-is-asymmetric-between-nix-and-opam.md
gares: I've seen this failure begin intermittent. I was wondering if 8.19 should be allowed to fail. It would be nice if it was allowed to fail only when rocq fails to compile, but not when a problem occurs after the dependencies are built.

shilangyu:
So the main goal of this PR is to cut CI times? I am not convinced making the setup this much more complicated is worth the reduction it achieves.

I would maybe instead question if running both the opam and nix versions is worth it? Since the end user gets what is produced by opam not nix, maybe we should be testing on opam installs instead?

gares:
We could also run OSX + Linux on recent rocq only, and just run linux-opam only on old rocq versions.
The python thing is too complex for me as well.

TDiazT:
> I'm not so sure I understand the approach from the description. What does "add a label to the repo mean"?

It's just having an additional label in the repo (ie. go to settings and add a new `ci-full` label) that you can set to the PR to trigger a full run

TDiazT:
> So the main goal of this PR is to cut CI times? I am not convinced making the setup this much more complicated is worth the reduction it achieves.

I mean, for me it was important to go from ~30 mins to ~10 minutes 😆 , even more if the mac part was only building and not doing anything else (and in so many versions).

> I would maybe instead question if running both the opam and nix versions is worth it? Since the end user gets what is produced by opam not nix, maybe we should be testing on opam installs instead?

Yeah, I agree, but I didn't know what the consensus was on this. 

> We could also run OSX + Linux on recent rocq only, and just run linux-opam only on old rocq versions.

I mean, we could still have it run completely in `main` (all rocq versions + OSX/Linux + opam/nix) but then in the PRs have it run as you say?
(This can also go behind the label if we need to trigger it in a PR just before merging)

I'm open to anything, I just felt there were too many jobs that made the (iterative) process too clunky.

gares:
I don't know what can be implemented is a simple way, but a workflow that we use (and like) in rocq is that by default PRs run a light-ci at each push, but cannot be merged without having run a full-ci, that can be requested by tagging a bot, that is also in charge of putting labels and refuses to merge if the labels are not set. I don't know if this is doable without such a bot. We could also ask the bot to be deployed here, but this CI thing is based on gitlab's ci, not sure can be ported easily to GH actions.

It would make sense to have a light ci that tests only the last rocq release and master (what one uses to develop) and have the full ci test older versions.

TDiazT:
Yeah, I thought about the bot used in Rocq, but I thought it was just simpler to go with a manual label for now. I was thinking on the same principle only that there was no automatic check on whether a full run was made, etc.

SkySkimmer:
>It would make sense to have a light ci that tests only the last rocq release and master (what one uses to develop) and have the full ci test older versions.

I think having the base CI test all supported rocq versions is useful. Seems likely that running full ci will get missed for overlays otherwise (if it's not run before the upstream PR is merged the needsfullci label will be ignored and the PR merged regardless)

PR #1305, opened 2026-08-25
ci: cancel superseded PR runs and bound every job with a timeout

TDiazT:
Added a concurrency group so a new push to a pull request cancels the run it supersedes, and a timeout-minutes to every job in place of the six-hour default (https://docs.github.com/en/actions/reference/limits). GitHub caps concurrent macOS runners at 5 while the two build routes schedule 16 macOS jobs, so a superseded run competes for the resource that is already scarce, and a job that hangs rather than fails holds a Darwin slot until someone notices. `cancel-in-progress` is limited to `pull_request`, because on main and on tags a cancelled run would leave a commit with no run of its own.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,12 +6,20 @@ on:
   pull_request:
   workflow_dispatch:
 
+# One run per ref: a second push to a PR cancels the first instead of stacking
+# another ~40 jobs behind it. Deliberately not on main or on tags, where
+# cancelling would leave a commit without a run of its own.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build-extension:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
     steps:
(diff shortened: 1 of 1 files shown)


gares:
This PR seems to do two things, correct me if I'm wrong.
One is to cancel outdated jobs, and this seems very useful, shame we did not set it up before.
The other is giving timeouts, which I'm not so sure I get. Can you clarify?

TDiazT:
> This PR seems to do two things, correct me if I'm wrong. One is to cancel outdated jobs, and this seems very useful, shame we did not set it up before. The other is giving timeouts, which I'm not so sure I get. Can you clarify?

Yeah, these could be two PRs. I can split them if you want.

The timeouts came from experimenting with some things and adding new tests, where the macOS test job would end up lingering because it was looping (it's a whole other problem...), and it would just continue a bit indefinitely (capped with the 6h anyway). Since there's already the issue of max 5 concurrent macOS jobs, this was taking all the space basically. So I figured it was a simple extra bonus we could add without much cost.

To decide the timeouts I had Claude get some estimates of previous runs and set a reasonable amount, that had some extra window in case a job was taking longer. The slowest thing I saw was install-opam on macOS at ~12 min against a 30 limit, so there's room. We can revisit them if you want.

gares:
ok, then let's keep these values and see how it goes

gares (approved):

PR #1306, opened 2026-08-25
ci: bump setup-node to v7 and let it cache the npm store

TDiazT:
Bumped setup-node from v3 to v7 across `ci.yml` and `publish-extension.yml` and turned on `cache: npm`, pointing cache-dependency-path at `client/package-lock.json` since the lockfile is not at the repository root. 

v3 was last released in April 2025, still declares the node16 action runtime the runner now overrides
(https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/), and bundles the form-data advisory CVE-2025-7783 that the v5 line fixed, on a runner holding a GITHUB_TOKEN. 

The cache skips downloading the 557 packages but still unpacks them, so it saves little until an install starts from an empty node_modules.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,11 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v6
     - name: Install Node.js
-      uses: actions/setup-node@v3
+      uses: actions/setup-node@v7
       with:
         node-version: 22.x
+        cache: npm
+        cache-dependency-path: client/package-lock.json
     - run: |
         cd client
         npm run package
@@ -122,10 +124,12 @@ jobs:
         echo "vsrocqtop=`which vsrocqtop`" >> $GITHUB_ENV
         vsrocqtop -v
 
-    - uses: actions/setup-node@v3
+    - uses: actions/setup-node@v7
       if: runner.os == 'Linux'
       with:
         node-version: 22
+        cache: npm
+        cache-dependency-path: client/package-lock.json
 
     - run: cd client && npm run package
       if: runner.os == 'Linux'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -16,9 +16,11 @@ jobs:
...
     - name: Install Node.js
-      uses: actions/setup-node@v3
+      uses: actions/setup-node@v7
       with:
         node-version: 22.x
+        cache: npm
+        cache-dependency-path: client/package-lock.json
     - run: |
         cd client
         npm run package
```
shilangyu: I think we should set the working directory for the entire job and then remove the `cd client` and the `cache-dependency-path`.

gares (approved):

.github/workflows/publish-extension.yml
```diff
@@ -24,9 +24,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
       - name: Install Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v7
         with:
           node-version: 22.x
+          cache: npm
+          cache-dependency-path: client/package-lock.json
```
shilangyu, line 34: Same here

shilangyu (changes requested):

.github/workflows/ci.yml
TDiazT: I'm not sure we can remove the `cache-dependency-path`. In the [doc](https://github.com/actions/setup-node#caching-global-packages-data) it says that:
>The action defaults to search for the dependency file (package-lock.json, npm-shrinkwrap.json or yarn.lock) in the repository root

There's a section on "Caching npm dependencies in monorepos" where they do this same thing. I don't think that changing the working directory actually applies to the root of the repo (we could try.. but I don't think it will work)

shilangyu: Ah shame. I forgot working-directory is only for `run` steps. It had a lesser effect than expected.

shilangyu (approved):

PR #1307, opened 2026-08-25
ci: pin every third-party action to a commit SHA

TDiazT:
Replaced the floating tag on all seventeen third-party uses with the commit SHA it resolves to today, keeping the release in a trailing comment, because a git tag is mutable and `avsm/setup-ocaml@v3` moved on 2026-08-20 in every job of all four workflows with nothing here recording it (https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). 

Each SHA is what its tag points at now, so nothing changes behaviour. `actions/checkout` and `actions/setup-node` stay on tags, being GitHub's own, and nothing moves off a stale major, which matters most for `softprops/action-gh-release` whose v1 tag still points at the v0.1.15 commit from 2022.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -16,7 +16,7 @@ jobs:
       uses: actions/checkout@v6
 
     - name: Use OCaml ${{ matrix.ocaml-compiler }}
-      uses: avsm/setup-ocaml@v3
+      uses: avsm/setup-ocaml@f92e0606b7ae4873dd1238465ea4bf6f8e40d85c  # v3.7.2
       with:
         ocaml-compiler: ${{ matrix.ocaml-compiler }}
 
@@ -24,9 +24,9 @@ jobs:
       run: opam install -y -j 2 opam-publish=2.5.0
 
     - name: Inject slug/short variables
-      uses: rlespinasse/github-slug-action@v4
+      uses: rlespinasse/github-slug-action@797d68864753cbceedc271349d402da4590e6302  # v4.5.0
 
-    - uses: webfactory/ssh-agent@v0.8.0
+    - uses: webfactory/ssh-agent@d4b9b8ff72958532804b70bbe600ad43b36d5f2e  # v0.8.0
       with:
           ssh-private-key: ${{ secrets.BOT_SSH_KEY }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
         (lang dune 3.2)
         (display short)
         EOF
-    - uses: cachix/install-nix-action@v31
+    - uses: cachix/install-nix-action@13d8dd58da0234aa297dedd986986ccb8e7f3e24  # v31.11.1
       with:
         nix_path: nixpkgs=channel:nixos-unstable
     - name: Check each yarn.nix up to date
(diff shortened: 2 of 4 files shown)


SkySkimmer:
Do we not want the auto update though? AFAIK CI has no secrets so security isn't a concern

gares:
I'm not a fan of this practice. If setup-ocaml made a dubious tag update once, it does not mean we do not trust them. I did not read the code the hash points to, to begin with. Maybe the tag update was exactly to fix a very nasty issue...

TDiazT:
My bad, I didn't mean to open this PR before talking with the team because I'm not that aware of the security part and permissions of the repo (I also have another PR adding more explicit permissions to jobs but that one I didn't open for the same reason).

> Do we not want the auto update though?

I agree we do, although pinning to a tag only means we're getting the latest of that major release, e.g., the `softprops/action-gh-release` was set to v1 and it's on v3 now, so we are not really getting the auto update. Same with `actions/setup-node` from v3 to v7.

Maybe we could add dependabot to actually bump the SHA to a newer version on a fixed period of time?
Then we could keep having the "latest" version.
I see that dependabot can update the SHA pins (and the `# vXXX` comments next to it), but the alerts don't work on SHA pins apparently ([docs](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts)). 
So maybe pinning does cost us vulnerability alerts, if they are enabled (are they?).

> AFAIK CI has no secrets so security isn't a concern

This might be me misreading, but I see that `cd.yml` and `publish-server.yml` have `BOT_SSH_KEY` and `OPAM_PUBLISH_TOKEN`. `publish-extension.yml` has `OVSX_PAT` and `VSCE_PAT`. All of these are relevant for publishing, right? Let me know if I'm reading them incorrectly.

For `ci.yml`,  I can't check the repo setting myself, but the `create-release` job calls `action-gh-release` without passing a `token:` input, so it's probably using the default GITHUB_TOKEN, and it does manage to create releases, which I think means that token has `contents: write`. If that's not an issue for you, we could maybe leave `ci.yml` unpinned and only pin the ones that publish?

TDiazT:
> If setup-ocaml made a dubious tag update once, it does not mean we do not trust them.

Yeah, it's true. Although `setup-ocaml` was one of the things that triggered this change for me, along with `setup-node` (no real update from v3->v7).
In `cd.yml` and `publish-server.yml` there's `avsm/setup-ocaml@v3` but in `ci.yml` there's `ocaml/setup-ocaml@v3` and I didn't understand why, and it turns out that avsm is pointing to the ocaml one and nothing in the workflow file tells me that. I couldn't find when the transfer happened either.

For me it's not that we do not trust them but it's more about being precise about what we are using.

> did not read the code the hash points to, to begin with.

I wouldn't either probably haha. Although the point for me is also more about having reproducible builds and not having unseen changes (like the `avsm` case). Same as having `package-lock.json` for the npm installs 🤷 .

gares:
OK, lest us start by fixing the mess (different versions of actions in different pipelines) and then decide if/how we want to pin them to a more precise version.

gares:
OK, lest us start by fixing the mess (different versions of actions in different pipelines) and then decide if/how we want to pin them to a more precise version.

SkySkimmer:
>This might be me misreading, but I see that cd.yml and publish-server.yml have BOT_SSH_KEY and OPAM_PUBLISH_TOKEN.

my mistake then

>Although the point for me is also more about having reproducible builds

setup-ocaml will run opam though so IDK if reproducible builds are possible

TDiazT:
> setup-ocaml will run opam though so IDK if reproducible builds are possible

Up to some point then I guess 😆

Issue #1286, opened 2026-07-21
Remove "vsrocq.proof.delegation"


MrQubo:
It looks like this option is known to be broken. Maybe it's better to remove it (as an user-facing setting), or at least add in the description that is has known problems?

There's this reported issue, where the comment says that this option is known to be broken:
https://github.com/rocq-prover/vsrocq/issues/940

I've also had a problem with it:
[#Rocq users > ✔ Qed "Attempt to save an incomplete proof"](https://rocq-prover.zulipchat.com/#narrow/channel/237977-Rocq-users/topic/.E2.9C.94.20Qed.20.22Attempt.20to.20save.20an.20incomplete.20proof.22/with/611978149)

TDiazT:
I was debugging the delegation issue with Claude and got this answer. I'm not familiar with this part of the code yet so I'm not sure how much sense it makes. Two apparently small bugs.
1. Delegate: the worker is never started when the delegated Qed is the last task
`ExecutionManager.execute` for a `PDelegate` does everything right: runs `interp_qed_delayed`, pushes the job onto `st.jobs`, and returns the `worker_available` Sel event. The consumer drops it. `checkingManager.ml:511-530` (post_execute):
```ocaml
match tasks with
| [] -> execution_finished st id started block_events (* 'events' not passed *)
| task :: tasks -> ... events = proof_view_event @ inject_em_events events @ ...
```
So whenever the delegated proof is the last task of the observetarget, the fork never happens. That is the end of file in Continuous mode, and stepping onto a `Qed` in Manual mode, which makes it look universally broken. If anything follows the Qed, delegation works end to end today: fork, worker runs the steps, results come back, diagnostics are correct, and the overview settles in ~0.13 s. Evidence:

| fixture (`workers: 1`, continuous, `block: false`) | Qed is last task | worker | settles | diagnostics |
|---|---|---|---|---|
| `Lemma x : True. Proof. idtac. exact I. Qed.` | yes | never starts | no, `(0,1,4)` | 0 |
| same + `Check x.` | no | forks, 3 steps | 0.135 s | 1, correct |
| `delegate_proof.v` + `Check x.` | no | forks, 4 steps | 0.132 s | 3, all correct |
| #940's `time auto.` + `Check x.` | no | forks | 0.255 s | correct, no crash |

[...]

Issue #940, opened 2024-11-05
`time` tactic breaks proof delegation


1024th:
A mininal example:

```coq
Lemma test: True.
Proof.
  time auto.
Qed.
```

If `vscoq.proof.delegation` is set to:
- "No delegation (all proofs are checked one after the other)", then things go well;
- "Skip the proofs which are out of focus. This requires a continuous checking to be off": then VsCoq complains `(in proof test): Attempt to save an incomplete proof`;
- "Delegate proofs": then `vscoqtop` crashes (dumps core).

gares:
We are aware that skip/delegation is currently broken. Thanks for opening an issue for the problem

PR #1308, opened 2026-08-26
test: isolate the client tests from one another

TDiazT:
The tests share one VS Code instance and one server, and nothing is reset between them, so both the published diagnostics and the settings carry over from one test to the other.

Diagnostics are keyed by URI and the server does not retract them when a document is closed (`textDocumentDidClose` in `lspManager.ml` emits no events),  so a test might end up reading the diagnostics that a previous test produced. `openFixture` now gives each test a copy at a URI of its own,  and the teardown deletes the copies.

Configuration updates were not awaited, which allowed a fixture to be checked with the previous setting.
They are also written to `testFixture/.vscode/settings.json`, which survived the run, so a run that set `proof.delegation` to `Delegate` left the next one starting from `Delegate`.
`configure` now awaits them and the teardown removes them.

Ran into this issue while trying to set up some other tests and changing the 10s sleep.
Specifically with the delegation test (`Delegating proofs`) that was passing only because the `Skipping proofs` test was leaving diagnostics behind on the same path.
When isolated it times out because under the correct settings (`proof.delegation: "Delegate"`) the server doesn't publish any diagnostics.
I see that  in #940 and #1286 there's mentions of delegation being broken, so the test is skipped for now.

### DIFF
--- a/client/src/test/suite/common.ts
+++ b/client/src/test/suite/common.ts
@@ -5,13 +5,72 @@ import * as path from "node:path";
 import * as vscode from "vscode";
 // import * as myExtension from '../../extension';
 
-export async function openTextFile(file: string): Promise<vscode.Uri> {
-    const docUri = vscode.Uri.file(
-        path.resolve(__dirname, "../../../testFixture", file),
+const fixtureRoot = path.resolve(__dirname, "../../../testFixture");
+
+let copiesMade = 0;
+const copies: vscode.Uri[] = [];
+const settingsChanged = new Set<string>();
+
+/**
+ * Copies a fixture to a URI of its own and opens that copy.
+ *
+ * Every test shares one VS Code instance and one server, and diagnostics are
+ * keyed by URI. The server does not retract them when a document is closed
+ * (`lspManager.ml`'s `textDocumentDidClose` emits no events), so a test that
+ * opens the same fixture as an earlier one reads the earlier one's results and
+ * cannot tell them from its own. Opening a fresh copy is what makes a test
+ * observe only what it caused.
+ *
+ * The copies live in the fixture directory, so they are checked under the same
+ * workspace root as the originals, and `resetTestState` deletes them.
+ */
+export async function openFixture(fixture: string): Promise<vscode.Uri> {
+    const { name, ext } = path.parse(fixture);
+    const copy = vscode.Uri.file(
+        path.join(fixtureRoot, `${name}_${copiesMade++}${ext}`),
     );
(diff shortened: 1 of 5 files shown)


gares:
This seems a really nice change, but I'm pretty bad at reading the ts code. Maybe @shilangyu or @KacperFKorban have a minute to review this.

gares:
can you rebase and see how CI is now?

client/src/test/suite/common.ts
```diff
@@ -5,13 +5,72 @@ import * as path from "node:path";
...
 // import * as myExtension from '../../extension';
 
-export async function openTextFile(file: string): Promise<vscode.Uri> {
-    const docUri = vscode.Uri.file(
-        path.resolve(__dirname, "../../../testFixture", file),
+const fixtureRoot = path.resolve(__dirname, "../../../testFixture");
+
+let copiesMade = 0;
+const copies: vscode.Uri[] = [];
+const settingsChanged = new Set<string>();
```
shilangyu, line 12: I don't like that there are global variables and that we need to keep track of things to then undo them afterwards.

Is it too expensive to just have a separate instance for every test instead? I can't imagine this solution being the expected one for E2E tests.

client/src/test/suite/common.ts
```diff
@@ -5,13 +5,72 @@ import * as path from "node:path";
...
+ * out. `update` is asynchronous, so an unawaited call leaves a fixture free to
+ * be opened and checked under the previous configuration.
+ */
+export async function configure(
+    section: string,
+    value: unknown,
+): Promise<void> {
+    settingsChanged.add(section);
+    await vscode.workspace.getConfiguration().update(section, value);
+}
```
shilangyu, line 55: Please use `client/src/configuration.ts` to strongly type this function. Otherwise nothing will catch us changing the names/types of settings.

client/src/test/suite/common.ts
TDiazT: Why not? I agree if it's production code but with tests I'm honestly ok (with the variables).

I can try and see if it's too expensive to launch a new instance each time, but I imagine it's comparatively a lot more. For just a couple of global variables (in tests) I don't know if the tradeoff is worth it

TDiazT: Each test takes ~5s (4s to launch + 1s to run all 4 tests), so changing it would be from 5s to 20s (plus any additional tests that are added later).
I removed the global state and changed it for a class that keeps track of common context, which is then setup/teardown in every test, so there's no conflicts.

PR #1309, opened 2026-08-28
Remove folding range setting

shilangyu:


gares:
@shilangyu I'm all for merging this one, but it seems to rever also changes to package.lock which I believe are not needed. Can you regenerate the lock file?

language-server/vsrocqtop/lspManager.ml
```diff
@@ -451,10 +448,7 @@ let documentFoldingRange id params =
   | None -> log (fun () -> "[documentFoldingRange] ignoring event on non existent document"); Error({message="Document does not exist"; code=None})
   | Some { st } ->
     log (fun () -> "[documentFoldingRange] getting folding ranges");
-    let raw_document = Dm.DocumentManager.Internal.raw_document st in
-    if Dm.RawDocument.line_count raw_document > !folding_line_count_cut_off then
-      Ok None
-    else if Dm.DocumentManager.is_parsing st then
+    if Dm.DocumentManager.is_parsing st then
```
gares, line 451: Should we keep returning Ok None so to not show the user a error?

shilangyu: I think its the same problem in every other place where we return an error. Do you know if there is some way to wait for the parsing to finish? Ideally we would schedule a response to this request whenever parsing is done

gares: Sure, but I was thinking at the bug you found in the folding-range PR where being to slow was giving a user visible error popup rather than just aborting the request. Wasn't it fixed by returning `Ok None` here rather than `Error ...`?

I just want to be sure we do not reintroduce the bug you found back then

gares: > Ideally we would schedule a response to this request whenever parsing is done

I don't think we have such a machinery ready to be used, but it is surely implementable. About parsing, I'd rather work on the PR that limits the portion of document being re-parsed.

shilangyu: > Sure, but I was thinking at the bug you found in the folding-range PR where being to slow was giving a user visible error popup rather than just aborting the request.

That error popup is going to happen for all other requests where we return an error if parsing is not done. So if an error popup would appear here, it would also appear in other places. I don't think this introduces any regressions to what we had before.

PR #1322, opened 2026-09-15
fix: two logging defects in completionSuggester and document

TDiazT:
Just small details. One is adding an extra newline and the other was just missing Printf


shilangyu (approved):
CI failure unrelated, but should be investigated separately

PR #1324, opened 2026-09-17
build: Keep dune's default flags on the executables

TDiazT:
This should bring back a bunch of warnings that were not showing up because `(flags -linkall)` replaces dune's `:standard` flags instead of extending them. (I now saw a couple of unused functions, variables, etc.)


shilangyu (approved):
CI failure unrelated.

PR #1327, opened 2026-09-22
Ajout des mots-clés "SProp" et "Abbreviation"

alex44119:


TheoWinterhalter:
Perhaps you should communicate in English. :)

